PR #1, opened 2019-04-01
add initial README content

zlozano:
Generated repo content using SDCLI, and added initial project overview.


zlozano:
> To clarify, is the scope of this system to collect _all_ network available compute assets or is it limited is some way such as "only AWS resources"?

While the first iteration will only support cloud assets, I don't think I've limited the working to just AWS assets. I refer mostly to "network assets". I do however, use AWS as an example use case. Should I add more use cases which are not cloud related?

PR #2, opened 2019-04-01
Secd 409/post api stub

zlozano:
https://github.com/asecurityteam/asset-inventory-api/pull/1 is a prereq to this PR.

This PR adds a stubbed out implemented of the POST API. This will enable consumers to be able to begin work as we continue to make progress on this project.


api.yaml
```diff
@@ -80,16 +81,70 @@ paths:
...
         lambda:
-          arn: "name_or_arn"
+          arn: "insert"
           async: false
-          request: '{}'
-          success: '{"status": 200, "bodyPassthrough": true}'
-          error: '{"status": 500, "bodyPassthrough": true}'
+          request: '#! json .Request.Body !#'
+          success: '{"status": 201, "bodyPassthrough": true}'
+          error: '{"status": #! if eq .Response.Body.errorType "InvalidInput" !# 400 #! else !# 500 #! end !#, "bodyPassthrough": true}'
```
bkrebsbach: Nit: could you break this into multiple lines so it's a little easier to read?

pkg/handlers/v1/cloud_insert.go
```diff
@@ -0,0 +1,68 @@
...
+		assetChanges.Changes = append(assetChanges.Changes, domain.NetworkChanges{
+			PrivateIPAddresses: val.PrivateIPAddresses,
+			PublicIPAddresses:  val.PublicIPAddresses,
+			Hostnames:          val.Hostnames,
+			ChangeType:         val.ChangeType,
+		})
+	}
+	if e := h.Storage.StoreCloudAsset(ctx, assetChanges); e != nil {
+		logger.Error(logs.StorageError{Reason: e.Error()})
+		return e
```
bkrebsbach: I like the InvalidInput error type. Given that we're toggling some behavior in transportd with error types, should this return a StorageError error type?

zlozano: The reason for switching on the error type was to inform the response code. This type of error will be a 500, so I don't see an immediate need to give it its own type since it doesn't require special handling in the serverfull-gateway response.

PR #4, opened 2019-04-09
database abstraction layer

mikerott:
As of commit https://github.com/asecurityteam/asset-inventory-api/pull/4/commits/b5b6870d2e9123b0bfe5786ae7b14995a55b4cf1, a full run of the `DB` struct and its functions have been tested, by (beware, [fish shell](https://fishshell.com/) syntax):
```
set -x POSTGRES_HOSTNAME localhost
# etc...
go build integration_test_db.go
./integration_test_db
# output is good, but inspected the DB tables anyway
```
Where `integration_test_db.go` used the settings lib, initialized the DB connection, ran a few functions, including queries, and printed output.  Full integration test writeup is forthcoming in a near future PR.


zlozano:
Looks like there are some linter errors: https://travis-ci.com/asecurityteam/asset-inventory-api/builds/107663673

integration_test_db.go
```diff
@@ -0,0 +1,15 @@
+package main
```
zlozano: Did you mean to include this file?

pkg/handlers/v1/db.go
```diff
@@ -0,0 +1,326 @@
+package v1
```
zlozano: This should be moved to a package named after the file in which the interface is defined. In this case `storage`
https://developer.atlassian.com/platform/security/guides/google-golang/service-repo/

pkg/handlers/v1/db.go
```diff
@@ -0,0 +1,326 @@
...
+	"time"
+
+	"github.com/asecurityteam/asset-inventory-api/pkg/domain"
+	"github.com/asecurityteam/asset-inventory-api/pkg/logs"
+
+	_ "github.com/lib/pq" // postgres driver for sql must be imported so sql finds and uses it
+)
+
+/*
+CREATE TABLE aws_resources (id VARCHAR PRIMARY KEY, account_id VARCHAR NOT NULL, region VARCHAR NOT NULL, resource_meta JSONB);
```
zlozano: This might be a good this to put in docs.go

main.go
```diff
@@ -5,32 +5,33 @@ import (
...
-}
-
 func main() {
 	ctx := context.Background()
+
+	db := v1.DB{}
+	err := db.Init(ctx); err != nil {
+		panic(err.Error())
+	}
+	if err != nil {
```
bkrebsbach: This block appears to be redundant from line 20

pkg/handlers/v1/db.go
```diff
@@ -0,0 +1,326 @@
...
+	"aws_ips " +
+	"FULL JOIN aws_ips_hostnames " +
+	"ON aws_ips.ip = aws_ips_hostnames.aws_ips_ip " +
+	"AND aws_ips.ts = aws_ips_hostnames.aws_ips_ts " +
+	"FULL JOIN aws_hostnames " +
+	"ON aws_hostnames.hostname = aws_ips_hostnames.aws_hostnames_hostname " +
+	"AND aws_hostnames.ts = aws_ips_hostnames.aws_hostnames_ts " +
+	"WHERE "
+
+var once sync.Once
```
bkrebsbach: This should be an unexported field of the DB struct

pkg/logs/db_info.go
```diff
@@ -0,0 +1,7 @@
+package logs
+
+// DBInfo is logged when there is some info to log with the backend dependency
+type DBInfo struct {
+	Message string `logevent:"message,default=database-info"`
```
kconwayatlassian: We try to make our log messages as specific as possible so that they are identifiable more easily in the log output. My recommendation to most folks is to use these rules for operational data:

- Logging at less than an error level indicates a lack of confidence that the code functions correctly. Convert DEBUG and INFO logs to tests.
- For every unrecoverable exception there is a metric that contributes to an alert/alarm.
- For every unrecoverable exception there is one, and only one, error log that provides the context into that state.
- For every error log emitted by a service there is one, and only one, location within the code from which that log text originates.
- All related logs have a single identifier by which they may be isolated when viewing.

None of these are true in all situations and am happy to chat about them in more depth. I share them more to illustrate that we've preferred more specific logging wherever possible. For example, instead of having a `DBError` we might have `DBConnectionFailed` or instead of `DBInfo` we might have `DBConnectionSucceeded` where each has their own, specific message. We've found it helps a lot during incidents.

pkg/domain/cloud.go
```diff
@@ -12,7 +15,40 @@ type CloudAssetChanges struct {
...
+type InboundResourceTags map[string]interface{}
+
+// Value implements the driver.Valuer interface to marshal InboundResourceTags to []byte
+func (p InboundResourceTags) Value() (driver.Value, error) {
+	j, err := json.Marshal(p)
+	return j, err
+}
+
+// Scan implements the sql.Scanner interface to unmarshal the database data into a map, which is the same type as received and written into the database
+func (p *InboundResourceTags) Scan(src interface{}) error {
```
zlozano: I don't think this is the best place for this functionality. The intent of the domain package is to declare domain types that are used throughout the various packages in the project. This functions seems specific to the data layer only.

pkg/domain/cloud.go
```diff
@@ -12,7 +15,40 @@ type CloudAssetChanges struct {
 	AccountID    string
 	Region       string
 	ResourceID   string
-	Tags         map[string]string
+	ARN          string
+	Tags         InboundResourceTags
```
zlozano: Didn't we determined this would be a flat string->string map?

pkg/domain/config.go
```diff
@@ -0,0 +1,40 @@
+package domain
+
+import (
+	"context"
+)
+
+// PostgresConfig contains the Postgres database configuration arguments
+type PostgresConfig struct {
+	Hostname     string `json:"hostname"`
```
zlozano: I don't think the json tags are correct here. Looking at the [docs](https://github.com/asecurityteam/settings), it looks like this should be a description tag with a short description of each variable.

pkg/domain/config.go
```diff
@@ -0,0 +1,40 @@
+package domain
```
zlozano: This should probably go in the same package as the db implementation, similar to https://github.com/asecurityteam/awsconfig-filterd/blob/master/pkg/filter/resource.go

pkg/domain/config.go
```diff
@@ -0,0 +1,40 @@
...
+// and may be used by the settings.NewComponent function.
+type PostgresConfigComponent struct{}
+
+// Settings populates a set of defaults if none are provided via config.
+func (*PostgresConfigComponent) Settings() *PostgresConfig {
+	return &PostgresConfig{}
+}
+
+// New constructs a PostgresConfig from a config.
+func (*PostgresConfigComponent) New(_ context.Context, c *PostgresConfig) (*PostgresConfig, error) {
```
zlozano: I am not _super_ familiar with the settings library, but I think this should return the thing that uses the configuration, in this case, the DB abstraction.

pkg/handlers/v1/cloud_insert.go
```diff
@@ -16,6 +16,7 @@ type CloudAssetChanges struct {
 	AccountID    string            `json:"accountId"`
 	Region       string            `json:"region"`
 	ResourceID   string            `json:"resourceId"`
+	ARN          string            `json:"arn"`
```
zlozano: ResourceID should be unique enough. What was the reason for adding this bit of information?

If we do want to use this, we will require changes to the upstream services, the SH schema, and the docs. Could you create a ticket or tickets to address this?

pkg/handlers/v1/storage/db.go
```diff
@@ -0,0 +1,309 @@
+package storage
```
zlozano: This needs to be moved out from under the handlers/v1 package, and under the pkg folder. https://developer.atlassian.com/platform/security/guides/google-golang/service-repo/

pkg/handlers/v1/storage/db.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	return nil
+}
+
+func (db *DB) saveResource(ctx context.Context, cloudAssetChanges domain.CloudAssetChanges) error {
+	// You won't get an ID back if nothing was done.  Also, this lib won't return the ID anyway even without the "ON CONFLICT DO NOTHING".
+	// See https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql
+	sqlStatement := fmt.Sprintf(`INSERT INTO %s VALUES ($1, $2, $3, $4, $5) ON CONFLICT DO NOTHING RETURNING id`, tableAWSResources)  // nolint
+
+	if _, err := db.sqldb.ExecContext(ctx, sqlStatement, cloudAssetChanges.ARN, cloudAssetChanges.AccountID, cloudAssetChanges.Region, cloudAssetChanges.ResourceType, cloudAssetChanges.Tags); err != nil {
```
zlozano: does this not need to be a part of the transaction as well?

pkg/handlers/v1/storage/db.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// StoreCloudAsset an implementation of the Storage interface that records to a database
+func (db *DB) StoreCloudAsset(ctx context.Context, cloudAssetChanges domain.CloudAssetChanges) error {
+	tx, err := db.sqldb.Begin()
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
zlozano: if `Commit()` or `Rollback()` return an error, we will miss it. If either return with an error, we'd want to return it from `StoreCloudAsset`. I'd recommend looking at the [example](https://play.golang.org/p/nDjr7Z-wSNR) usage from the golang docs. Though instead of emitting a Fatal log, it may be better to return the original error if the rollback was successful, otherwise [wrap](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) the original error inside the rollback error.

pkg/domain/cloud.go
mikerott: Yes but in order for marshalling/unmarshalling, you need a type so you can implement functions for the sql lib to call

pkg/domain/config.go
mikerott: OK. I wasn't sure about this one

pkg/handlers/v1/storage/db.go
mikerott: OK but I thought you said earlier this should be in a package named for what it implements..?

pkg/handlers/v1/storage/db.go
mikerott: Yes it does. Good catch

pkg/domain/cloud.go
zlozano: Okay. In that case we should probably keep that encapsulated somewhere in the data layer/storage package and no in the model used between packages.

pkg/handlers/v1/storage/db.go
mikerott: OK will improve it

pkg/handlers/v1/storage/db.go
zlozano: Yep that's still true. So instead of `pkg/handlers/v1/storage` it should be `pkg/storage`

pkg/handlers/v1/cloud_insert.go
mikerott: I could be convinced either way.  Ben and I both thought ARN would be better, and I think he's already working on the other services to pass it through. We should discuss..

pkg/domain/config.go
mikerott: That doesn't feel right.. We should discuss though

pkg/handlers/v1/cloud_insert.go
zlozano: I don't have any strong opinions either way as longs as all the necessary things get updated

pkg/domain/config.go
mikerott: That's probably me being lazy.. I don't think we need json here... at least not yet. I'll take them out

pkg/domain/config.go
zlozano: Okay sounds good. I was making that call based off of https://github.com/asecurityteam/awsconfig-filterd/blob/master/pkg/filter/resource.go

and the components listed here: https://github.com/asecurityteam/transportd/tree/master/pkg/components

pkg/domain/cloud.go
mikerott: I'd have to make a data model for the DB INSERT, which will be very redundant with the existing data model, but I can do it for encapsulation

zlozano: I don't think you need to remake the whole model, do you? just a new type you can convert the tags into

pkg/domain/config.go
mikerott: Yeah I saw his Ben did that. It confused me for a bit... I think the component/config impls should be just that and no more.  In other words, no factories or domain struct creation in there, but you know more go than me. Learn me

pkg/domain/cloud.go
mikerott: Maybe but that new type would be outside of the domain package, yet still referenced from within this struct, which is in the domain package.  Me encapsulation would start leakin

mikerott: Actually I'll remove this and change it back to `map[string]string`.  I can encapsulate the conversion in the DB struct functions, which is probably where it really belongs anyway.  (this addresses your next comment too)

pkg/domain/config.go
zlozano: > no factories

But I think that is the intended purpose of settings.

```
// All configuration is driven by a struct that uses standard go types.
type Config struct {}
// The resulting element is virtually anything that you need it to be
// for the purposes of the system you are building.
type Result struct {}
// The component contract interfaces between the thing you want to make,
// the result, and the settings project. It is responsible for producing
// instances of the configuration struct with any default values populated.
// It is also used to construct new instances of the result using a
// populated configuration. No references to the settings project are
// required to create any part of this setup.
type Component struct {}
func (*Component) Settings() *Config { return &Config{} }
func (*Component) New(_ context.Context, c *Config) (*Result, error) {
    return &Result{}, nil
```

main.go
```diff
@@ -4,37 +4,42 @@ import (
...
-}
-
 func main() {
 	ctx := context.Background()
+
 	insert := &v1.CloudInsertHandler{
 		LogFn:   runhttp.LoggerFromContext,
 		StatFn:  runhttp.StatFromContext,
-		Storage: &nopStorage{},
+		Storage: &db,
```
zlozano: `db` hasn't been defined at this point. Since main.go isn't included in the unit tests, I usually do a `go build` and a linter check before I push.

main.go
```diff
@@ -4,37 +4,42 @@ import (
...
 	}
+
+	postgresConfigComponent := &storage.PostgresConfigComponent{}
+	postgresSettings := new(storage.PostgresSettings)
+	err = settings.NewComponent(ctx, source, postgresConfigComponent, postgresSettings)
+	if err != nil {
+		panic(err.Error())
+	}
+	db := storage.DB{}
+	if err := db.Init(ctx); err != nil {
```
zlozano: Your `Init()` function looks to take a configuration struct as well.

pkg/handlers/v1/cloud_insert.go
zlozano: Did we end up making tickets for this work?

pkg/storage/config.go
```diff
@@ -0,0 +1,40 @@
...
+// and may be used by the settings.NewComponent function.
+type PostgresConfigComponent struct{}
+
+// Settings populates a set of defaults if none are provided via config.
+func (*PostgresConfigComponent) Settings() *PostgresConfig {
+	return &PostgresConfig{}
+}
+
+// New constructs a PostgresConfig from a config.
+func (*PostgresConfigComponent) New(_ context.Context, c *PostgresConfig) (*PostgresConfig, error) {
```
zlozano: Based on the [docs](https://github.com/asecurityteam/settings) I still don't think this is the correct usage of the settings Factory. Given: 

```
// All configuration is driven by a struct that uses standard go types.
type Config struct {}
// The resulting element is virtually anything that you need it to be
// for the purposes of the system you are building.
type Result struct {}
// The component contract interfaces between the thing you want to make,
// the result, and the settings project. It is responsible for producing
// instances of the configuration struct with any default values populated.
// It is also used to construct new instances of the result using a
// populated configuration. No references to the settings project are
// required to create any part of this setup.
type Component struct {}
func (*Component) Settings() *Config { return &Config{} }
func (*Component) New(_ context.Context, c *Config) (*Result, error) {
    return &Result{}, nil
}
```

Our `DB` struct should take the place of the `Result` struct. @kconwayatlassian can you provide some guidance on the usage here?

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+	tx, err := db.sqldb.Begin()
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		switch err {
+		case nil:
+			if commitErr := tx.Commit(); commitErr != nil {
+				log.Fatalf(`Failed to commit transation: %s`, commitErr)
```
zlozano: We should not log here, and we should almost never log at the Fatal level. Fatal will call `os.Exit()` and cause our app to terminate. This should be pulled out of the defer so that we can return these errors from the top level function.

pkg/domain/config.go
kconwayatlassian: I think what @zlozano is saying is that `json` tags _won't_ be used here since this only interfaces with the settings loader and it has a different set of tags for adding metadata.

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	return nil
+}
+
+func (db *DB) insertNetworkChangeEvent(ctx context.Context, timestamp time.Time, isPublic bool, isJoin bool, resourceID string, ipAddress string, hostname *string, tx *sql.Tx) error {
+	// Postgres does not auto-create partition tables, so we do it.  We're using 3-month intervals (quarters)
+	monthInterval := 3
+	year := timestamp.Year()
+	fromMonth := (int((timestamp.Month())-1)/monthInterval)*monthInterval + 1 // get the interval of the year we're in, then use the first month of that quarter
```
zlozano: unnecessary parens around `(timestamp.Month())`

pkg/domain/config.go
kconwayatlassian: Yeah, `settings` components are effectively factories, or factory factories, that get interface with a special configuration loader.

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	defer rows.Close()
+
+	networkChangeEvents := make([]domain.NetworkChangeEvent, 0)
+	for rows.Next() {
+		var row domain.NetworkChangeEvent
+		var bytes []byte
+		if err = rows.Scan(&row.ResourceID, &row.IPAddress, &row.Hostname, &row.IsPublic, &row.IsJoin, &row.Timestamp, &row.AccountID, &row.Region, &row.Type, &bytes); err != nil {
+			log.Fatal(err)
```
zlozano: remove log.Fatal (see earlier comment)

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+			return nil, err
+		}
+		if bytes != nil {
+			var i map[string]string
+			json.Unmarshal(bytes, &i) // we already checked for nil, and the DB column is JSONB; no need for err check here
+			row.Tags = i
+		}
+		networkChangeEvents = append(networkChangeEvents, row)
+	}
+
```
zlozano: I think we should check `rows.Err()` here

pkg/domain/network_change_event.go
```diff
@@ -0,0 +1,55 @@
...
+	"encoding/json"
+	"errors"
+	"time"
+)
+
+// NetworkChangeEvent represent a result row of database query
+type NetworkChangeEvent struct {
+	ResourceID string
+	IPAddress  string
+	Hostname   sql.NullString
```
kconwayatlassian: Echoing @zlozano , bundling the implementation details with the domain type leaks the abstraction more than we typically allow. Our usual pattern for this is to stick to standard lib types (as much as possible) for domain types and have the implementation specific details hidden by using unexported types in the implementation packages. For example:

```golang
// domain/something.go
type Change struct {
 Hostname *string
}
```

```golang
// someting/pgsql.go
type pgsqlChange struct{
 Hostname sql.NullString
}
func (s *PgsqlSomething) StoreChange(ctx context.Context, ch Change) error {
  pgch := pgsqlChange{
    Change: sql.NullString(ch.Hostname),
  },
  // ...
}
```

It adds a small layer of tedium to adapt the types but, largely, helps us prevent making implementation specific domain types that have to change every time we change our underlying data model.

pkg/handlers/v1/storage/db.go
kconwayatlassian: The general rule is to name the implementation package after the domain type that it implements. We also have a soft rule around naming the files containing an implementation after the concrete's name. For example, if given a domain type of ```type Storage interface{}``` then we'd most expect to see `/pkg/storage/pgsql.go` or `/pgk/storage/redis.go`.

pkg/handlers/v1/storage/db.go
```diff
@@ -0,0 +1,309 @@
...
+	for _, hostname := range changes.Hostnames {
+		err := db.insertHostname(ctx, hostname, tx)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, val := range changes.PrivateIPAddresses {
+		isJoin := false
+		if changes.ChangeType == "ADDED" {
```
kconwayatlassian: For all cases of string comparison, I'd recommend using a constant value for the sentinel and using https://golang.org/pkg/strings/?m=all#EqualFold rather than equality to ensure that we account for input that isn't cased the way we expect.

pkg/handlers/v1/storage/db.go
```diff
@@ -0,0 +1,309 @@
...
+	fromMonth := (int((timestamp.Month())-1)/monthInterval)*monthInterval + 1 // get the interval of the year we're in, then use the first month of that quarter
+	fromDay := 1
+	toMonth := (int((timestamp.Month())-1)/monthInterval)*monthInterval + monthInterval        // get the interval of the year we're in, then use the last month of that interval
+	toDay := time.Date(year, time.Month(toMonth+1), 0, 0, 0, 0, 0, timestamp.Location()).Day() // time.Date will normalize that toMonth + 1 and 0 day shenanigans
+	from := fmt.Sprintf(`%d-%02d-%02d`, year, fromMonth, fromDay)
+	to := fmt.Sprintf(`%d-%02d-%02d`, year, toMonth, toDay)
+	partitionTableNameSuffix := fmt.Sprintf(`%d_%02dto%02d`, year, fromMonth, toMonth)
+	tableName := fmt.Sprintf(`%s_%s`, tableAWSEventsIPSHostnames, partitionTableNameSuffix)
+
+	if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s PARTITION OF %s FOR `+  // nolint
```
kconwayatlassian: I don't have any data around this but my naive assumption would be that this is an expensive check to run on _every_ insert. Additionally, when this actually results in table creation then my naive expectation is that this is likely an order of magnitude or more slower than an insert especially when factoring in the index creation. Would it be possible to automate this on a schedule instead?

pkg/storage/config.go
kconwayatlassian: I'll copy my previous comment since it was lost in the change:
`settings` components are effectively factories, or factory factories, that get interfaced with a special configuration loader. The `Settings()` method should return the configuration struct with any default values set and the `New` method should return the actual thing being constructed. In this case, I suggested we use a component because it enables us to make the concrete storage implementation selectable and configurable via configuration. The output should be an implementation of whatever interface represents the storage backend.

pkg/storage/db.go
kconwayatlassian: To extend this, we don't use the standard library logger anywhere because it does not enable use to use structured logging. Anything that would be logged should be logged using the logger from context and with a struct. This helps ensures that all of our logs are unique, contain the appropriate context, and can be marshalled to a structured output for log ingestion systems.

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+	fromMonth := (int((timestamp.Month())-1)/monthInterval)*monthInterval + 1 // get the interval of the year we're in, then use the first month of that quarter
+	fromDay := 1
+	toMonth := (int((timestamp.Month())-1)/monthInterval)*monthInterval + monthInterval        // get the interval of the year we're in, then use the last month of that interval
+	toDay := time.Date(year, time.Month(toMonth+1), 0, 0, 0, 0, 0, timestamp.Location()).Day() // time.Date will normalize that toMonth + 1 and 0 day shenanigans
+	from := fmt.Sprintf(`%d-%02d-%02d`, year, fromMonth, fromDay)
+	to := fmt.Sprintf(`%d-%02d-%02d`, year, toMonth, toDay)
+	partitionTableNameSuffix := fmt.Sprintf(`%d_%02dto%02d`, year, fromMonth, toMonth)
+	tableName := fmt.Sprintf(`%s_%s`, tableAWSEventsIPSHostnames, partitionTableNameSuffix)
+
+	if _, err := tx.ExecContext(ctx, fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s PARTITION OF %s FOR `+ // nolint
```
kconwayatlassian: Copying over comments that were lost in the change:
I don't have any data around this but my naive assumption would be that this is an expensive check to run on every insert. Additionally, when this actually results in table creation then my naive expectation is that this is likely an order of magnitude or more slower than an insert especially when factoring in the index creation. Would it be possible to automate this on a schedule instead?

pkg/storage/db.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, hostname := range changes.Hostnames {
+		err := db.insertHostname(ctx, hostname, tx)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, val := range changes.PrivateIPAddresses {
+		isJoin := false
+		if changes.ChangeType == "ADDED" {
```
kconwayatlassian: Copying again:
For all cases of string comparison, I'd recommend using a constant value for the sentinel and using https://golang.org/pkg/strings/?m=all#EqualFold rather than equality to ensure that we account for input that isn't cased the way we expect.

pkg/domain/network_change_event.go
mikerott: Oh I didn't realize I left that in there.  I'll take it out.

main.go
mikerott: I thought for sure I had done that, but maybe it was just my local integration test.  I _know_ I did the lint, but it didn't catch it.  Anyway... my bad.

pkg/storage/config.go
mikerott: That's fine.  I'll change my impl to reflect the intent of the Settings/Component lib.

pkg/storage/db.go
mikerott: I see.  Will fix.

pkg/storage/db.go
mikerott: Not sure.  It _is_ in a transaction, so not sure if it's really expensive or not.  That said, I _really_ don't like that it's here on every run INSERT.  There are several mechanisms to optimize.  One I was thinking of was just an in-memory array of "known-to-already-exist tables" so we don't bother with that `CREATE`.  That in-memory array could arguably be considered "state", but I don't think it really matters in this case.  I'll impl it, and you guys can comment.

zlozano: In the interest of time and size of this PR, this could potentially be something that we ticket out and address after we merge this code in. We don't _have_ to do it that way. It's only a suggestion.

kconwayatlassian: I'm OK with splitting it out to reduce the PR size. In case we continue it here:

The table cache will still result in a cache miss for the first request to each running instance as well as the first request to each running instance every time the clock changes to a new partition-able segment of time.

For the occasions where a new table _is_ created, it's not clear to me how the transaction will affect the query. For example, if we have 5 concurrent writes occurring on each of our 3 running instances then will all 3 instances block until the table create is finished or will two of them fail the transaction? If they block then we become subject to queue based latency as our landing rate for new writes should remain predictable during that time but all writes become enqueued. Alternatively, how would this kind of `DDL in a transaction` work if someone is running this system on a multi-primary configuration of postgresql?

I'd feel much more comfortable if there were separate path we could execute on-demand and in advanced to perform the DDL needed to create a new partition table. This would enable operators to more strictly control this process as well as help mitigate much of the risk of performing dynamic DDL.

pkg/storage/db.go
```diff
@@ -0,0 +1,283 @@
...
+func (db *DB) StoreCloudAsset(ctx context.Context, cloudAssetChanges domain.CloudAssetChanges) error {
+
+	logger := runhttp.LoggerFromContext(ctx)
+
+	tx, err := db.sqldb.Begin()
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
kconwayatlassian: This `defer` is guaranteed to run before the function returns a value. We've never used it as most cases end up seeming like anti-patterns but there is the option of defining the method with a named return value. For example: https://play.golang.org/p/9ScRVbnxw_W .

I'd be happy if we never used named return values but SQL style transactions may be one place where they fit. Normally I'd rather have a log in non-handler code than named returns but the current state of this method is that it could return `nil` but actually fail.

I'm OK starting with this so long as we track it and figure out what we want to do long term before going to production.

pkg/storage/db.go
```diff
@@ -0,0 +1,283 @@
...
+			row.Tags = i
+		}
+		if hostname.Valid {
+			row.Hostname = hostname.String
+
+		}
+		networkChangeEvents = append(networkChangeEvents, row)
+	}
+
+	return networkChangeEvents, nil
```
kconwayatlassian: This is still missing `rows.Err()` in the return.

pkg/storage/db_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+	ctx := context.Background()
+	ctx = logevent.NewContext(ctx, stdoutLogger(ctx))
+
+	if err = thedb.StoreCloudAsset(ctx, fakeCloudAssetChanges()); err == nil {
+		t.Errorf("was expecting an error, but there was none")
+	}
+	assert.Equal(t, "could not start transaction", err.Error())
+
+	if err := mock.ExpectationsWereMet(); err != nil {
```
kconwayatlassian: `testify` has `assert.NotError()` and `require.NotError` as helpers for these cases.

pkg/storage/db.go
zlozano: poke -- I wish GH had tasks :sigh:

pkg/handlers/v1/cloud_insert.go
mikerott: No, but I'll follow up.

pkg/storage/db.go
```diff
@@ -0,0 +1,283 @@
...
+			sslmode := "disable"
+			if host != "localhost" {
+				sslmode = "enable"
+			}
+			psqlInfo := fmt.Sprintf("host=%s port=%s user=%s "+
+				"password=%s dbname=%s sslmode=%s",
+				host, port, user, password, dbname, sslmode)
+			pgdb, err := sql.Open("postgres", psqlInfo)
+			if err != nil {
+				initerr = err
```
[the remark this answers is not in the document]
mikerott: I'll fix it.

pkg/storage/db.go
mikerott: I'm gonna do what Zach suggested in the next comment.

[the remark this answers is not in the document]
mikerott: Will do this.

pkg/storage/db.go
mikerott: The thing I might worry about is the timestamp indicated by the AWS config change report being currently processed.  I _think_ we can safely assume that all reports we get have accurate near-real-time timestamps, but we should be careful that we pre-create the partition tables if we're not creating them inline as I've first coded it.  Also, we'd want to code for if/when we need to retrieve or replay anything from cold/old storage.

I'll open up a performance ticket; we need to also avoid constantly running unnecessary INSERTs into the aws_resources table too.

pkg/handlers/v1/cloud_insert.go
mikerott: SECD-481 opened

pkg/storage/db.go
mikerott: Opened SECD-482

Gopkg.toml
```diff
@@ -28,3 +28,14 @@
...
+[[constraint]]
+  name = "github.com/DATA-DOG/go-sqlmock"
+  version = "1.3.3"
+
+[[constraint]]
+  name = "github.com/lib/pq"
+  version = "v1.0.0"
+
+[[constraint]]
+  name = "github.com/lib/errors"
```
kconwayatlassian: FYI, https://godoc.org/golang.org/x/exp/errors is incubating which _should_ inter-operate with Dave's pkg/errors. Given that, it might be time to start making use of this library more standard for our projects. We haven't used it historically in favor of sticking to standard library support. However, the combined features of pkg/errors and x/errors would make a lot of the kinds of error handling we do a bit easier and less tedious.

pkg/storage/db.go
```diff
@@ -0,0 +1,280 @@
...
+	networkChangeEvents := make([]domain.NetworkChangeEvent, 0)
+	for rows.Next() {
+		var row domain.NetworkChangeEvent
+		var bytes []byte
+		var hostname sql.NullString
+		err = rows.Scan(&row.ResourceID, &row.IPAddress, &hostname, &row.IsPublic, &row.IsJoin, &row.Timestamp, &row.AccountID, &row.Region, &row.Type, &bytes)
+		if err == nil {
+			if bytes != nil {
+				var i map[string]string
+				json.Unmarshal(bytes, &i) // nolint: we already checked for nil, and the DB column is JSONB; no need for err check here
```
kconwayatlassian: nit: the "idiomatic" way to express this condition and silence the errcheck linter is with `_ = json.Unmarshal()`. The underscore character is a special one in go that can be used anywhere we want to actively ignore a value.

pkg/storage/db.go
kconwayatlassian: nit: This could be reduced to `return networkChangeEvents, rows.Err()`. There's not obligation to ensure the left-hand value is a zero value when an error is returned. Unless otherwise documented, such as https://golang.org/pkg/io/#Reader, the general expectation is that a non-nil error means the caller must assume that any other values returned are invalid.

pkg/storage/config_test.go
```diff
@@ -0,0 +1,40 @@
...
+	postgresConfigComponent := PostgresConfigComponent{}
+	postgresConfig := postgresConfigComponent.Settings()
+	assert.NotNil(t, postgresConfig)
+	assert.Empty(t, postgresConfig.DatabaseName)
+}
+
+func TestShouldMakeNewDB(t *testing.T) {
+	called := false
+	postgresConfig := PostgresConfig{}
+	originalDBInitFn := dbInitFn
```
kconwayatlassian: Doing this creates a constraint whereby we can never run tests in parallel because this relies on global state. Why aren't we able to use the sql mock here that is used elsewhere in the test instead?

pkg/storage/db.go
```diff
@@ -21,6 +21,9 @@ const tableAWSEventsIPSHostnames = "aws_events_ips_hostnames"
 
 const added = "ADDED" // one of the network event types we track
 
+// test seam
+var sqlOpenFn = sql.Open
```
kconwayatlassian: Same question as with the component test since global state is one of the things we typically don't allow and it limits the ways we can run the tests.

pkg/storage/config_test.go
```diff
@@ -34,7 +34,7 @@ func TestShouldMakeNewDB(t *testing.T) {
 	}
 
 	postgresConfigComponent := PostgresConfigComponent{}
-	postgresConfigComponent.New(context.Background(), &postgresConfig)
+	postgresConfigComponent.New(context.Background(), &postgresConfig) // nolint
```
kconwayatlassian: Rather than `//nolint`, can we have `_, _ = postgresConfigComponent.New(...)` to explicitly ignore the error?

pkg/storage/config_test.go
mikerott: sqlmock does not offer a mock for `Open`.  I could, though, put a reference to the function in the struct rather than making it a global.  Yeah?

pkg/storage/db.go
mikerott: See comment above..

PR #5, opened 2019-04-09
SECD-448 Add GET API Stub

bkrebsbach:
Add a stubbed out GET API for fetching cloud assets.


api.yaml
```diff
@@ -140,6 +209,39 @@ components:
...
+          items:
+            type: string
+        hostnames:
+          type: array
+          items:
+            type: string
+        createdAt:
+          type: string
+          format: date-time
+        deletedAt:
```
kconwayatlassian: What are createdAt and deletedAt intended to convey here? Are they proxies for when a system came online or went offline?

api.yaml
```diff
@@ -89,6 +89,75 @@ paths:
             #! if eq .Response.Body.errorType "InvalidInput" !# 400
             #! else !# 500
             #! end !#, "bodyPassthrough": true}'
+  /cloud:
+    get:
+      summary: "Retrieve a cloud asset"
+      parameters:
+        - name: "ipAddress"
```
kconwayatlassian: We need to be careful with this API. Any query param may be given more than once but we don't define a behavior for this. Additionally, it's not clear from the specification if these conditions are combined with `OR` or `AND` or something else.

I'm also hesitant to offer one API that makes all attributes queryable. This quickly degenerates into "implement an API that allows consumers to run arbitrary SQL statements in the database". I'd much prefer to start with a very limited set of operations targeted at attribution for specific details of the asset. This might be, for example, an API of `/cloud/ip/{ipAddress}?time=XXX`. I'm open to different ways of shaping the API but I do think we should start with heavy restrictions on what we support in terms of search complexity.

pkg/domain/storage.go
```diff
@@ -2,9 +2,11 @@ package domain
 
 import (
 	"context"
+	"time"
 )
 
 // Storage interface provides functions for inserting and retrieving assets
 type Storage interface {
 	StoreCloudAsset(context.Context, CloudAssetChanges) error
+	FetchCloudAsset(ctx context.Context, hostname string, ipAddress string, timestamp time.Time) (CloudAssetDetails, error)
```
kconwayatlassian: Similar to my API comments. I think we should start with highly targeted interfaces for attributes rather than the "catch-all". As the number of searchable values for an asset grows, this interface will become unwieldy. At least in the beginning, I'd rather see things like:
```golang
type IPFetcherBy interface {
   FetchByIP(ctx context.Context, when time.Time, ipAddress string) (CloudAssetDetails, error)
}
```
This will both enable us to focus only on implementing the queries from a single perspective (rather than exposing a dynamic query function for the entire database) as well as enable new features to be developed without modifying one, monolithic feature function. Again, I'm OK drifting towards a more complex/powerful system later but let's keep it constrained for now.

main.go
```diff
@@ -27,8 +39,14 @@ func main() {
 		StatFn:  runhttp.StatFromContext,
 		Storage: &nopStorage{},
 	}
+	fetch := &v1.CloudFetchHandler{
+		LogFn:   runhttp.LoggerFromContext,
```
kconwayatlassian: For the sake of consistency, I'd recommend having an alias in `domain` that defines these context functions like we do for the `LogFn` and `Log` types. This way there aren't direct references to `runhttp` outside of `domain`.

bkrebsbach: I'll replace the usage of `runhttp` with `domain` aliases.

pkg/domain/storage.go
bkrebsbach: Does each query need its own interface, or would it be sufficient to separate queries on each attribute to separate functions within a single interface? For example, splitting the `Storage` interface into something like:

```go
type CloudAssetStorer interface {
    Store(context.Context, CloudAssetChanges) error
}

type CloudAssetFetcher interface {
    FetchByIP(ctx context.Context, when time.Time, ipAddress string) (CloudAssetDetails, error)
    FetchByHostname(ctx context.Context, when time.Time, hostname string) (CloudAssetDetails, error)
}
```

api.yaml
bkrebsbach: These have been removed.

api.yaml
bkrebsbach: I've updated the API to match the more targeted `/cloud/ip/{ipAddress}?time=XXX` style

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -0,0 +1,145 @@
...
+func (h *CloudFetchByIPHandler) Handle(ctx context.Context, input CloudAssetFetchByIPParameters) (CloudAssets, error) {
+	logger := h.LogFn(ctx)
+
+	ts, e := time.Parse(time.RFC3339Nano, input.Timestamp)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return CloudAssets{}, InvalidInput{Field: "time", Cause: e}
+	}
+
+	if input.IPAddress == "" {
```
kconwayatlassian: Assuming input validation fails or is disabled, is there any harm in making a query for `ipAddress == ""`? I'm not sure if this is actually a bad request or a not found condition.

bkrebsbach: I think that would depend on how the underlying storage mechanism handles a query for an empty string. It'd be putting unnecessary load on the database - the real question would be how much.

That said, this should be rejected by input validation though, so I think it makes sense to remove it.

main.go
```diff
@@ -15,20 +15,56 @@ import (
 
 type nopStorage struct{}
 
-func (s *nopStorage) StoreCloudAsset(ctx context.Context, _ domain.CloudAssetChanges) error {
-	runhttp.LoggerFromContext(ctx).Info("store cloud asset stub")
+func (s *nopStorage) Store(ctx context.Context, _ domain.CloudAssetChanges) error {
```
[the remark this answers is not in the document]
bkrebsbach: Fixed.

main.go
```diff
@@ -15,20 +15,56 @@ import (
...
+	domain.LoggerFromContext(ctx).Info("store cloud asset stub")
```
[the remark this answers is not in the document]
bkrebsbach: Fixed.

PR #6, opened 2019-04-19
Feature/db integration tests

mikerott:
### work-in-progress (WIP)

April 19 7:00am US Central; if you're reviewing, review the overall technique (locations of sql files, use of packr, use of TestMain and the utility functions in storage_test.go), not yet the actual tests; those are still being worked

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -31,20 +31,36 @@ const commonQuery = "SELECT " +
 	"WHERE "
 
 // can't use Sprintf in a const, so...
-// %s throughout should be `aws_hostnames_hostname` or `aws_ips_ip`
-const latestStatusQuery = "SELECT aws_events_ips_hostnames.aws_resources_id, aws_events_ips_hostnames.aws_ips_ip, aws_events_ips_hostnames.aws_hostnames_hostname, is_public, is_join, aws_events_ips_hostnames.ts, aws_resources.account_id, aws_resources.region, aws_resources.type, aws_resources.meta " +
-	"FROM " +
-	"    (SELECT %s, aws_resources_id, max(ts) as ts " +
+// %s should be `aws_hostnames_hostname` or `aws_ips_ip`
+const latestStatusQuery = "WITH latest_candidates AS ( " +
+	"    SELECT " +
+	"        *, " +
+	"        MAX(ts) OVER (PARTITION BY aws_events_ips_hostnames.aws_resources_id) as max_ts " +
 	"    FROM aws_events_ips_hostnames " +
-	"    WHERE ts BETWEEN $2 AND $3 " +
-	"    GROUP BY %s, aws_resources_id " +
-	") AS latest " +
-	"    JOIN aws_events_ips_hostnames " +
-	"    ON latest.ts = aws_events_ips_hostnames.ts " +
-	"    JOIN aws_resources " +
-	"    ON aws_events_ips_hostnames.aws_resources_id = aws_resources.id " +
-	"WHERE latest.%s = aws_events_ips_hostnames.%s " +
-	"AND latest.%s = $1;"
+	"    WHERE " +
+	"        aws_events_ips_hostnames.%s = $1 AND " +
+	"        aws_events_ips_hostnames.ts <= $2 " +
+	"), " +
+	"latest AS ( " +
+	"    SELECT * " +
+	"    FROM latest_candidates " +
+	"    WHERE latest_candidates.ts = latest_candidates.max_ts " +
+	") " +
+	"SELECT " +
+	"    latest.aws_resources_id, " +
+	"    latest.aws_ips_ip, " +
+	"    latest.aws_hostnames_hostname, " +
+	"    latest.is_public, " +
+	"    latest.is_join, " +
+	"    latest.ts, " +
+	"    aws_resources.account_id, " +
+	"    aws_resources.region, " +
+	"    aws_resources.type, " +
+	"    aws_resources.meta " +
+	"FROM latest " +
+	"    LEFT OUTER JOIN " +
+	"    aws_resources ON " +
+	"        latest.aws_resources_id = aws_resources.id;"
 
 // DB represents a convenient database abstraction layer
 type DB struct {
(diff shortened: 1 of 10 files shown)


tests/storage_test.go
```diff
@@ -0,0 +1,260 @@
...
+		t.Fatalf("Failed to DROP tables due to: %s", err.Error())
+	}
+	if err := createTables(db); err != nil {
+		t.Fatalf("Failed to CREATE tables due to: %s", err.Error())
+	}
+}
+
+func dropTables(db *sql.DB) error {
+
+	sqlFile := "1_clean.sql"
```
zlozano: I don't see this file

tests/storage_test.go
```diff
@@ -0,0 +1,260 @@
...
+	if _, err = db.Exec(s); err != nil {
+		return err
+	}
+	fmt.Println(s)
+	return nil
+}
+
+func createTables(db *sql.DB) error {
+
+	// fmt.Printf("DROPping existing aws_* tables\n")
```
zlozano: Will this eventually be restored? Or is it dead code?

tests/storage_test.go
```diff
@@ -0,0 +1,260 @@
...
+	if err = settings.NewComponent(ctx, source, postgresConfigComponent, dbStorage); err != nil {
+		panic(err.Error())
+	}
+
+	db, err = connectToDB()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	// if err = dropTables(db); err != nil {
```
zlozano: will these if blocks be used eventually?

pkg/storage/db.go
```diff
@@ -30,6 +30,22 @@ const commonQuery = "SELECT " +
 	"	ON aws_resources_id = aws_resources.id " +
 	"WHERE "
 
+// can't use Sprintf in a const, so...
+// %s throughout should be `aws_hostnames_hostname` or `aws_ips_ip`
+const latestStatusQuery = "SELECT aws_events_ips_hostnames.aws_resources_id, aws_events_ips_hostnames.aws_ips_ip, aws_events_ips_hostnames.aws_hostnames_hostname, is_public, is_join, aws_events_ips_hostnames.ts, aws_resources.account_id, aws_resources.region, aws_resources.type, aws_resources.meta " +
```
kconwayatlassian: I'd leave it to the optimizer/scheduler to determine the truth of the matter, but I believe this query will result in several scans/nested loops because of the organization. I believe we can avoid them by restructuring the query with common table expressions and window functions. For example:

```sql
// Assuming:
// $1 = hostname
// $2 = timestamp requested
//
// Compute a filtered set of results that each has an attached "most recent ts"
// that is up to but not exceeding the given timestamp.
WITH latest_candidates AS (
    SELECT
        *,
        // the max is applied after the WHERE so the dataset to aggregate is much smaller.
        MAX(ts) OVER (PARTITION BY aws_events_ips_hostnames.aws_resources_id) as max_ts
    FROM aws_events_ips_hostnames
    WHERE
        aws_events_ips_hostnames.hostname = $1 AND
        aws_events_ips_hostnames.ts <= $2
),
// Drop all but only the most recent matching records for each resource
// that would have been the truth at the requested time.
latest AS (
    SELECT *
    FROM latest_candidates
    WHERE latest_candidates.ts = latest_candidates.max_ts
)
SELECT
    latest.*,
    aws_resources.account_id,
    aws_resources.region,
    aws_resources.type,
    aws_resources.meta
FROM latest
LEFT OUTER JOIN
    aws_resources ON 
        latest.aws_resources_id = aws_resources.id
```

I'm pulling from a fairly cold part of my SQL knowledge cache so the syntax may not be 100% correct. However, this is a general strategy I've seen work well for removing nested sub-queries and, often as a result, full table scans and nested loops from the optimizer plans. It also removes the need for a second timestamp by removing the need to do any kind of range query.

One thing that I'm definitely not sure about is how partitioning affects all of this.

pkg/storage/db.go
```diff
@@ -240,6 +256,22 @@ func (db *DB) GetIPAddressesForHostname(ctx context.Context, hostname string) ([
 	return db.runQuery(ctx, sqlstmt, hostname)
 }
 
+// GetAssetsByHostnameAtTime gets the assets who have hostname at the specified time
+func (db *DB) GetAssetsByHostnameAtTime(ctx context.Context, hostname string, at time.Time) ([]domain.NetworkChangeEvent, error) {
+	// ok to hardcode this timestamp as the service in which this code runs went into production sometime after this date:
+	from, _ := time.Parse(time.RFC3339, "2019-04-22T00:00:00+00:00")
```
kconwayatlassian: If we stick with a strategy like this, won't we be bound to always scan all partitions of a table between the epoch date and the search date? It seems like that would minimize the usefulness of a time based partition.

tests/storage_test.go
mikerott: It's been checked in now, in a squashed commit.

tests/storage_test.go
mikerott: I've deleted that comment.

pkg/storage/db.go
mikerott: It's a good question.  I suppose I could use `max(ts)`..  I'll think about it and try to re-work this strategy.

pkg/storage/db.go
mikerott: Great!  thanks!  I'll work through these improvements.

tests/storage_test.go
```diff
@@ -0,0 +1,592 @@
...
+
+	before(t, db)
+
+	// code should tolerate no data in the tables
+
+	from, _ := time.Parse(time.RFC3339, "2019-08-08T08:29:35+00:00")
+	to, _ := time.Parse(time.RFC3339, "2019-08-10T08:29:35+00:00")
+	networkChangeEvents, err := dbStorage.GetIPAddressesForTimeRange(ctx, from, to)
+	if err != nil {
+		t.Fatal(err.Error())
```
kconwayatlassian: This can be replaced with `require.NoError(t, err)`. `require` is the fatal version of `assert` where `assert` will actually continue the test but register the error.

pkg/storage/db.go
mikerott: Incorporated your suggested changes.  Thanks again!

PR #7, opened 2019-04-24
Feature/api to db integration

mikerott:


pkg/storage/db.go
```diff
@@ -254,37 +247,19 @@ func (db *DB) insertNetworkChangeEvent(ctx context.Context, timestamp time.Time,
...
-func (db *DB) GetAssetsByIPAddressAtTime(ctx context.Context, ipAddress string, at time.Time) ([]domain.NetworkChangeEvent, error) {
+// FetchByIP gets the assets who have IP address at the specified time
+func (db *DB) FetchByIP(ctx context.Context, when time.Time, ipAddress string) ([]domain.CloudAssetDetails, error) {
 	sqlstmt := fmt.Sprintf(latestStatusQuery, `aws_ips_ip`)
-	return db.runQuery(ctx, sqlstmt, ipAddress, at)
+	return db.runQuery(ctx, sqlstmt, ipAddress, when)
 }
 
-func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) ([]domain.NetworkChangeEvent, error) {
+func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) ([]domain.CloudAssetDetails, error) { //([]domain.NetworkChangeEvent, error) {
```
bkrebsbach: Nit: Looks like there's a leftover comment at the end of this line

pkg/storage/db.go
```diff
@@ -293,23 +268,66 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
 
 	defer rows.Close()
 
-	networkChangeEvents := make([]domain.NetworkChangeEvent, 0)
+	cloudAssetDetails := make([]domain.CloudAssetDetails, 0)
+
+	tempMap := make(map[string]*domain.CloudAssetDetails)
```
zlozano: Why a map of `*domain.CloudAssetDetails` rather than the value-type version?

pkg/storage/db.go
```diff
@@ -293,23 +268,66 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
...
+					}
+				}
+				if !found {
+					tempMap[row.ResourceID].Hostnames = append(tempMap[row.ResourceID].Hostnames, hostname.String)
+				}
+			}
+			found = false
+			if ipAddress.Valid {
+				var ipAddresses *[]string
+				if isPublic.Bool { // no need for .Valid check; DB table column guarantees non-null
```
zlozano: If I understand the comment, is it necessary to use `sql.NullBool` for this?

pkg/storage/db.go
```diff
@@ -293,23 +268,66 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
...
+
 		if err == nil {
-			if bytes != nil {
+			if metaBytes != nil {
 				var i map[string]string
-				_ = json.Unmarshal(bytes, &i) // we already checked for nil, and the DB column is JSONB; no need for err check here
+				_ = json.Unmarshal(metaBytes, &i) // we already checked for nil, and the DB column is JSONB; no need for err check here
 				row.Tags = i
 			}
+			if tempMap[row.ResourceID] == nil {
```
bkrebsbach: You might want to check with @zlozano to ensure there are no collisions with his work to replace `ResourceID` with `ARN`

zlozano: I'll wait for him to merge before I start my work on this service. The merge conflicts will be :toodamnhigh:

pkg/storage/db.go
mikerott: Mostly so I could do [this nil check and just assign the address](https://github.com/asecurityteam/asset-inventory-api/blob/feature/api-to-db-integration/pkg/storage/db.go#L293-L294) of the newly created `row` struct.

If I change it by making it a value rather than a pointer, like:
```
tempMap := make(map[string]domain.CloudAssetDetails)
```

Then I have to change the if clause linked above like so:
```
if _, existingValue := tempMap[row.ResourceID]; !existingValue {
	tempMap[row.ResourceID] = row
}
```

And I got more code to change later when the map is no longer a map of pointers, and [this SO question and answer explains](https://stackoverflow.com/questions/32751537/why-do-i-get-a-cannot-assign-error-when-setting-value-to-a-struct-as-a-value-i) it:

image

My uneducated preference would be to keep this code as-is, but if there really is a pitfall I'm gonna hit, of course I'll change it.

pkg/storage/db.go
mikerott: You're right, I didn't need that extra bit.  Good catch.

pkg/storage/db.go
zlozano: Ah okay. that makes sense 👍

PR #10, opened 2019-04-29
update to Postgres 11 to take advantage of automatic index propagation

mikerott:


scripts/2_create.sql
```diff
@@ -25,37 +25,36 @@ CREATE TABLE aws_hostnames
...
--- Postgres 11 has many updates for partitioning, and when we can use Postgres 11, we can uncomment
--- the FOREIGN KEY bits.  See:  https://pgdash.io/blog/partition-postgres-11.html
+-- And we'll make sure there's an index right away that Postgres 11 will automatically propagate to child tables:
+-- (We use aws_ips_ip and timestamp as it is expected these will be the most frequently used in queries)
 
--- CREATE INDEX
--- IF NOT EXISTS aws_events_ips_hostnames_aws_ips_ip_ts_idx ON aws_events_ips_hostnames USING BTREE
--- (aws_ips_ip, ts);
+CREATE INDEX
+IF NOT EXISTS aws_events_ips_hostnames_aws_ips_ip_ts_idx ON aws_events_ips_hostnames USING BTREE
```
zlozano: Request: Can we create an index on hostname/ts? I think vuln scanning team will be searching by this dimension as well.

kconwayatlassian: We shouldn't need all of those indexes on the join table. If we're querying by a specific dimension then we should query the table that records that dimension and then join it back to the details.

We really shouldn't need additional indexes on the join table (which is effectively a view at this point) because the primary keys and foreign keys are already indexed.  Additionally, the primary keys of the other fact and dimension tables are indexed. Adding additional indexes somewhere should be treated like a code smell in 4NF or 5NF schema design.

pkg/storage/db.go
```diff
@@ -120,6 +140,41 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+func (db *DB) doesDBExist(pgdb *sql.DB, dbName string) bool {
+	row := pgdb.QueryRow("SELECT datname FROM pg_catalog.pg_database WHERE lower(datname) = lower('" + dbName + "');") // nolint
+	if err := row.Scan(); err != nil {
+		return false
+	}
+
+	return true
+}
+
+func (db *DB) create(pgdb *sql.DB, name string) error {
```
zlozano: nit: it doesn't look like this needs to be a function of `db` since it's not dealing with any of that struct's internal state

pkg/storage/db.go
```diff
@@ -120,6 +140,41 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// use function's intent is to close the existing connection (pgdb parameter) and open
+// a new one against the desired psqlInfo connection string
+func (db *DB) use(pgdb *sql.DB, psqlInfo string) (*sql.DB, error) {
```
zlozano: nit: it doesn't look like this needs to be a function of `db` since it's not dealing with any of that struct's internal state

zlozano:
Looks like codecov is reporting no coverage for the most recent commit.

zlozano:
I hate writing tests just for the sake of getting coverage, but it is the bar we set. You should be able to get the remaining coverage by testing some of the [error scenarios](https://codecov.io/gh/asecurityteam/asset-inventory-api/compare/a3a1d97353c450f8405da5ec6cd14bbc1b293640...781efb389d2857341cb05e22e2ecbdb93559e411/diff)

scripts/2_create.sql
```diff
@@ -25,37 +25,40 @@ CREATE TABLE IF NOT EXISTS aws_hostnames
...
+(aws_ips_ip, ts);
+
+CREATE INDEX
+IF NOT EXISTS aws_events_ips_hostnames_aws_hostnames_hostname_ts_idx ON aws_events_ips_hostnames USING BTREE
+(aws_hostnames_hostname, ts);
```
kconwayatlassian: Suggestion: There are too many secondary indexes on table that should be all foreign keys. This seems to me like we're not querying the data correctly. For example, we should not be querying the event table by hostname. Instead, we should query the hostname table by hostname and join it to the events table to get the data we need. The same is true for IPs or any other dimension we query by.

I'll take another look at the schema to see what I've missed but we're treating this table like a fully denormalized view of the data. That's not the original intent, IIRC.

scripts/0_wipe.sql
```diff
@@ -0,0 +1 @@
+DROP DATABASE %s;
```
kconwayatlassian: Question: Why make this a Go `fmt` string rather than valid SQL? Also, why does this logic not fit in the `1_clean.sql` script?

scripts/2_create.sql
mikerott: Right, agree on the unnecessary indices.  I'll remove the `CREATE INDEX` and sub in a code comment.

pkg/storage/db.go
mikerott: I've changed the logic a bit so that it _is_ part of the internal `db` struct

pkg/storage/db.go
mikerott: I've changed the logic a bit so that it _is_ part of the internal `db` struct

scripts/0_wipe.sql
mikerott: > Why make this a Go fmt string rather than valid SQL?
I did not want to hard-code the database name here; the db name is from env var.  Also, the sql lib won't take `$1` in that spot, it only uses `$#` in SELECT (or equivalent) statements.

> why does this logic not fit in the 1_clean.sql script
By splitting up the `DROP DATABASE` from the `1_clean.sql`, the _first_ run of `DB.Init` follows a different code path than each subsequent invocation, achieving a wee bit of better coverage.  It increased my confidence, at least.

kconwayatlassian: I don't mind having it one place or the other. It's just different than the other scripts by effectively requiring a parameter to be given. This wouldn't work with the `RunScript()` method of the DB.

PR #12, opened 2019-04-29
Secd 300/partition creation

zlozano:
based on the suggestion here: https://severalnines.com/blog/guide-partitioning-data-postgresql the simple approach to creating future partitions is to do them out of band.  This PR creates a handler to create the next quarterly partition.

This is fairly large, it may be best to go commit-by-commit

- will wait for @mikerott to merge his PR and rebase with that
- initially I went with a handler that only created the next partition up, but modified to take an optional timestamp


api.yaml
```diff
@@ -209,6 +209,55 @@ paths:
...
+          - "requestvalidation"
+          - "responsevalidation"
+          - "lambda"
+        lambda:
+          arn: "createPartition"
+          async: false
+          request: >
+            {
+              "hostname": "#!.Request.URL.hostname!#",
+              "time": "#!index .Request.Query.time 0!#"
```
bkrebsbach: Should this be `.Request.Query.timestamp`?

api.yaml
```diff
@@ -239,7 +288,7 @@ components:
         - resourceType
         - accountId
         - region
-        - resourceId
+        - arn
```
bkrebsbach: I thought this was removed? Does this repo need to be rebased on master?

api.yaml
```diff
@@ -209,6 +209,55 @@ paths:
...
+          - "accesslog"
+          - "requestvalidation"
+          - "responsevalidation"
+          - "lambda"
+        lambda:
+          arn: "createPartition"
+          async: false
+          request: >
+            {
+              "hostname": "#!.Request.URL.hostname!#",
```
bkrebsbach: Nit: it doesn't look like `hostname` is used in the handler, this should probably be dropped.

zlozano: Oh yeah. this is bad copy/pasta

api.yaml
zlozano: I branch off of Mikes branch and I don't think he had updated with master. I will see if I can resolve this.

pkg/storage/db.go
```diff
@@ -120,6 +122,61 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+func (db *DB) GeneratePartition(ctx context.Context) error {
+	stmt := `SELECT tablename FROM pg_catalog.pg_tables WHERE tablename LIKE 'aws_events_ips_hostnames_%' ORDER BY tablename DESC LIMIT 1`
+	row := db.sqldb.QueryRowContext(ctx, stmt)
+	var latest string
+	err := row.Scan(&latest)
+	switch err {
+	case nil:
+	case sql.ErrNoRows:
+		now := db.now
+		if now == nil {
```
kconwayatlassian: Opinion: You don't have to do this here if you want to make use of the `sync.Once` to set the value.

pkg/storage/db.go
```diff
@@ -120,6 +122,61 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
 	return initerr
 }
 
+// GeneratePartition finds the latest partition, and generate the next partition based on the previous partions time range
+func (db *DB) GeneratePartition(ctx context.Context) error {
+	stmt := `SELECT tablename FROM pg_catalog.pg_tables WHERE tablename LIKE 'aws_events_ips_hostnames_%' ORDER BY tablename DESC LIMIT 1`
```
kconwayatlassian: Suggestion: This is a clever way to ensure the always fast-forward partitioning but you're paying a premium in code logic and table naming contracts for it to work. You might consider using the same approach as a lot of DB schema migration tools and create a table that tracks the metadata you need to manage the schema. For example, you might have:

```sql
CREATE TABLE partitions (
  ts timestamp
  partition_begin timestamp
  partition_duration interval
)
```
And then use it with something like:

```sql
SELECT * FROM partitions ORDER BY partition_begin DESC LIMIT 1
```
Which would give you back pre-structured data for all the time values you need to generate the next partition. You'd insert a record into the table in the same transaction as creating the new table.

You'd buy yourself a couple of benefits with this approach. One is that it will remove all of the table name string parsing code and replace it with native types. Another is that it would enable you to adjust the partition duration at any time via configuration because the durations of each partition are recorded in the metadata rather than assumed.

api.yaml
```diff
@@ -209,6 +209,56 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
+  /v1/partitions:
```
kconwayatlassian: Question: Do you consider this part of the v1 API contract? Does that mean that PostgreSQL is not actually a choice and/or that a requirement for future implementations must support a concept of partition?

zlozano: I thought about not having it versioned, but did this mostly for consistency.

>  requirement for future implementations must support a concept of partition

It may make more sense to not version this

kconwayatlassian: Alternatively, we might have an `/ops` section where these things can go. I imagine that we _would_ have a versioned API contract for the endpoints but only have endpoints available based on the DB installed. For example, we might offer `/ops/pgsql/v1/...` for these cases.

api.yaml
```diff
@@ -198,6 +198,64 @@ paths:
...
             }
           success: '{"status": 200, "bodyPassthrough": true}'
+          error: >
+            {
+              "status":
+              #! if eq .Response.Body.errorType "InvalidInput" !# 400,
+              #! else !#
+              #! if eq .Response.Body.errorType "NotFound" !# 404,
+              #! else !#
+              #! if eq .Response.Body.errorType "PartitionConflict" !# 409,
```
bkrebsbach: Task: this is in the wrong lambda section, should be down under `/ops/pgsql/v1/partitions`

zlozano: Oh dang. thanks for catching that

pkg/storage/db.go
```diff
@@ -13,13 +13,17 @@ import (
...
 
-const tableAWSResources = "aws_resources"
-const tableAWSIPS = "aws_ips"
-const tableAWSHostnames = "aws_hostnames"
-const tableAWSEventsIPSHostnames = "aws_events_ips_hostnames"
-const createScript = "2_create.sql"
-
-const added = "ADDED" // one of the network event types we track
+const (
+	defaultPartitionInterval   = 3 // months
```
kconwayatlassian: Question: Why use an integer of months rather than a `time.Duration` value?

pkg/storage/db.go
```diff
@@ -199,6 +208,79 @@ func (db *DB) ping() error {
 	return nil
 }
 
+// GeneratePartition finds the latest partition, and generate the next partition based on the previous partions time range
+func (db *DB) GeneratePartition(ctx context.Context) error {
+	stmt := "SELECT partition_end FROM partitions ORDER BY partition_begin DESC LIMIT 1"
```
kconwayatlassian: Suggestion: I'm reading this as `Give me the partition_end for the latest partition_create`. Granted, if we're preserving consistency of the data then we won't have a problem but it's more technically correct to ask `Give me the latest partition_end` if we are guaranteeing a fast-forward only model.

pkg/storage/db.go
```diff
@@ -199,6 +208,79 @@ func (db *DB) ping() error {
...
+	row := db.sqldb.QueryRowContext(ctx, stmt)
```
kconwayatlassian: Suggestion: There are a number of conditions that could result in concurrent executions of this path even though we're relying on an explicit trigger via a special endpoint. I'd recommend looking at some of the locking options (https://www.postgresql.org/docs/11/sql-lock.html) and putting all of the partition management work inside a transaction. This would enable you to bail early if unable to acquire the lock and let the other running call finish.

pkg/storage/db.go
```diff
@@ -199,6 +208,79 @@ func (db *DB) ping() error {
...
+		return fmt.Errorf("Invalid partition range: %s, %v", latest, err)
+	}
+
+	return db.GeneratePartitionWithTimestamp(ctx, latestTime)
+}
+
+// GeneratePartitionWithTimestamp generates the partition based on the given time
+func (db *DB) GeneratePartitionWithTimestamp(ctx context.Context, begin time.Time) error {
+	monthInterval := defaultPartitionInterval
+	begin = time.Date(begin.Year(), begin.Month(), 1, 0, 0, 0, 0, begin.Location()) // month granularity
```
kconwayatlassian: Suggestion: You can both simplify this and enable non-month precision by using https://golang.org/pkg/time/#Time.Truncate to adjust the time in conjunction with a `time.Duration` instead of an integer month value.

pkg/storage/db.go
```diff
@@ -199,6 +208,79 @@ func (db *DB) ping() error {
...
+	end := begin.AddDate(0, monthInterval, 0)
+	partitionTableNameSuffix := fmt.Sprintf(`%d_%02dto%d_%02d`, begin.Year(), begin.Month(), end.Year(), end.Month())
+	tableName := fmt.Sprintf(`%s_%s`, tableAWSEventsIPSHostnames, partitionTableNameSuffix)
+
+	tx, err := db.sqldb.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	beginFmt := begin.Format(time.RFC3339)
```
kconwayatlassian: Question: I was under the impression that the `pg` driver handled `time.Time` types. Is that not correct?

pkg/storage/db.go
zlozano: there is no month duration. the golang durations go up to an hour

pkg/storage/db.go
zlozano: We'd have to build up a duration to represent days and months, and this seemed simpler: https://stackoverflow.com/questions/36988681/time-time-round-to-day

While using the partition table will allow us to have dynamic partition intervals, we are only using the default partition interval for now. I will create a ticket to address this.

pkg/storage/db.go
zlozano: Yes it does. But in some places (i.e. partition creation) we can't use prepared statements, rather we must use fmt.Sprintf(). I have updated the prepared statements to use time.Time

pkg/storage/db.go
kconwayatlassian: Do these need to be calendar months? Is the goal to align the partitions with the calendar? For example, would it be OK to have a 90 day interval even though that won't map perfectly to calendar months?

pkg/storage/db.go
```diff
@@ -60,6 +64,7 @@ type DB struct {
 	sqldb   *sql.DB // this is a unit test seam
 	scripts func(name string) (string, error)
 	once    sync.Once
+	now     func() time.Time // unit test seam
```
kconwayatlassian: nit: I don't think we need to call these out since this pattern of dependency management is largely inspired by our testing practices.

pkg/storage/db.go
zlozano: > Do these need to be calendar months? 

Not necessarily. There were a few things I considered during these changes:
1. the original implementation used calendar months, and this ticket was only around automating the partition creation, so I am trying to keep the scope to that
2. most reference material online uses calendar months. I am not sure of any well-documented advantages.
3. if we do need to eventually need to go to a more granular partition scheme, it is much easier to become more granular, as opposed to transitioning from every X days to every X calendar months

PR #18, opened 2019-05-09
only create partitions when needed

mikerott:
Originally this PR was intended to change the partition table create technique from "use an externally triggered endpoint to create the needed partition" to "use an internally-coded timer task to create the needed partition".  However, through discussion, we want to make this codebase as lambda compatible as possible, so the externally triggered endpoint makes the most sense, rather than a long-lived internal timer task (and thus long lived connection to a DB, which is not lambda friendly).

So this PR is relatively small, and only changes the behavior such that creation of the _next_ needed partition only occurs when it's known to be needed, and within 3 days of it being needed.


pkg/storage/db.go
```diff
@@ -151,10 +152,35 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+	quit := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				if err := db.generatePartition(ctx); err != nil {
+					logger := domain.LoggerFromContext(ctx)
+					logger.Error(logs.StorageError{Reason: err.Error()})
+				}
+			case <-quit:
```
zlozano: It doesn't look like this channel is ever written to. Ideally, we would be able to hook into some sort of lifecycle event to start/stop this, where the start function would execute this code path and the stop function would write to the quit channel causing the timer to stop.

The scenario I am thinking of is bringing up a new stack, while the old stack still lingers for a while in case of a rollback. In that case, the old stack will still be running and could possible trigger an extra run of this timer. Ultimately, since the table is locked, it is benign. However, as it is currently written, I don't think this code path will ever be executed.

pkg/storage/db.go
```diff
@@ -151,10 +152,35 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+	// run every 12 hours, with the assumption that partition tables will never be
+	// this granular, thus partition tables are created well in advance of their need
+	ticker := time.NewTicker(12 * time.Hour)
+	quit := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				if err := db.generatePartition(ctx); err != nil {
+					logger := domain.LoggerFromContext(ctx)
```
zlozano: this is likely an event we will want to emit stats on as well.

pkg/storage/db.go
```diff
@@ -151,10 +152,35 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+					logger.Error(logs.StorageError{Reason: err.Error()})
```
zlozano: suggestion: error logs should be actionable. if a PartitionConflict error is returned, that would indicate we tried to create a partition, but one for the specified time already existed. We may want to log that type of error at an info level.

pkg/storage/db.go
```diff
@@ -230,16 +257,25 @@ func (db *DB) GeneratePartition(ctx context.Context) error {
...
+		return nil
+	}
+
 	return generatePartitionForTime(ctx, tx, db.now(), latestTime)
 }
 
-// GeneratePartitionWithTimestamp generates the partition based on the given time
-func (db *DB) GeneratePartitionWithTimestamp(ctx context.Context, begin time.Time) error {
+// generatePartitionWithTimestamp generates the partition based on the given time
+func (db *DB) generatePartitionWithTimestamp(ctx context.Context, begin time.Time) error {
```
zlozano: The integration tests use this. I am not sure why the CI didn't fail. Is `make integration` passing for you locally?

pkg/storage/db.go
```diff
@@ -230,16 +257,25 @@ func (db *DB) GeneratePartition(ctx context.Context) error {
...
+	beginTime, err := time.Parse(time.RFC3339, begin)
+	if err != nil {
+		return handleRollback(tx, fmt.Errorf("Invalid partition range: %s, %v", latest, err))
+	}
 	latestTime, err := time.Parse(time.RFC3339, latest)
 	if err != nil {
 		return handleRollback(tx, fmt.Errorf("Invalid partition range: %s, %v", latest, err))
 	}
 
+	if beginTime.After(db.now()) {
```
bkrebsbach: Nit: I'm finding the partition creation flow here a little hard to follow, I think adding some more comments throughout the code about how this should function might help make this easier to follow and reason about.

pkg/storage/db.go
```diff
@@ -151,10 +152,35 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
 		initerr = db.RunScript(ctx, createScript)
+
+		if initerr == nil {
+			db.startPartitionGeneratorTimer(ctx)
+		}
 	})
 	return initerr
 }
 
+func (db *DB) startPartitionGeneratorTimer(ctx context.Context) {
```
zlozano: Since the lifecycle of this process is different than that of the parent context, you should create a new context here and copy the values from the parent context. I think @kconwayatlassian wrote some [code](https://github.com/asecurityteam/serverfull/blob/master/pkg/handlers/v1/invoke.go#L25) recently to make the copy process pretty easy.

Also, since this is not spawning from a request, I don't think a logger or stater will have been injected. We will likely need to inject a logger/stater instance into the parent context from main.go. It will essentially be mimicking [this](https://github.com/asecurityteam/logevent/blob/b32737d8d51c86c030b126ef9028cb2bab259aa1/http/middleware.go#L24) behavior

pkg/storage/db.go
```diff
@@ -151,10 +152,35 @@ func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
...
+	// run every 12 hours, with the assumption that partition tables will never be
+	// this granular, thus partition tables are created well in advance of their need
+	ticker := time.NewTicker(12 * time.Hour)
```
kconwayatlassian: Question: There are a few edge cases we'll need to consider for a Lambda environment. Using the question prefix since I'm looking for your thoughts.

Official runtime docs: https://docs.aws.amazon.com/lambda/latest/dg/running-lambda-code.html

In the most degenerate case, Lambda will create a new VM to handle a request and then immediately pause that VM. Any VM re-use is not guaranteed and is only an optimization we may get. In the degenerate case, this timer will never fire because the VM doesn't exist more more than the milliseconds it takes to resolve one function call. In the case where a VM is reused, I _think_ the event will fire after looking at the Go code that manages time based events. 

Another thing to consider is that the instance count of this function is not under our control and _will_ be scaled out to meet demands on the function invocation.  The default limit is [1000 concurrent executions](https://docs.aws.amazon.com/lambda/latest/dg/limits.html) which I'm taking to mean 1000 instances of the function running in isolated VMs. We do have some control over this but by having this as a part of the DB connection init means that it will run as part of all instances of all functions we run. There are only three after this PR but we expect more to be added over time for different query types. For us, this might mean several hundred instances racing to lock the partition tables and run DDL statements. This might not be a problem but concurrent DDL in other systems has caused me problems in the past so I get nervous when I see it.

All of this is to say that Lambda isn't really intended to support this kind of behavior. What's the driving motivator to make this time based instead of trigger based?

If it helps, Lambda can be triggered on a [schedule via CloudWatch](https://docs.aws.amazon.com/lambda/latest/dg/tutorial-scheduled-events-schedule-expressions.html) and we have an internal cron-like system that can provide this functionality as well.

mikerott: Super good points.  I was not thinking about the execution conditions this code would find itself in under lambda.  Now that I _am_ thinking about it... :)

@zlozano and I talked over the endpoint trigger as the mechanism for ensuring pre-emptive existence of the required child partition table, and had some worries about it.  Specifically, how to ensure the cronman schedule was correct.

That said, given it's lambda, maybe we _should_ run cronman schedule daily to hit a trigger endpoint, and just let the service decide if it's really time to create the next partition or not.

I'm inclined to go back to that ^^^ trigger solution now.

mikerott: @kconwayatlassian and @zlozano and I talked, each independently, and both have great points.  We shall talk in person and record the result here.  Meantime, I'm going to improve test coverage.

pkg/storage/db.go
mikerott: I put some comments below, which did not update GH's comment ^^^.  /shrug

pkg/storage/db.go
mikerott: Result if conversation is recorded in the PR description.

PR #21, opened 2019-05-28
pass partition duration as a parameter

zlozano:
Currently, the database partitions are hardcoded to 3 months. This PR makes a couple of adjustments:

- Change the default partition from 3 months to 90 days
- allows callers to specify a custom partition duration in # of days

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	defaultPartitionInterval   = 3 // months
+	defaultPartitionInterval   = 90 // days
 	tablePartitions            = "partitions"
 	tableAWSResources          = "aws_resources"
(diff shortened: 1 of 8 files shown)


pkg/storage/db.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	defaultPartitionInterval   = 3 // months
+	defaultPartitionInterval   = 90 // days
```
kconwayatlassian: question: why not switch to the `time.Duration` type now that it's not fixed?

zlozano: I felt this was more readable than time.Hour * 2160. Also, we currently don't want any partial day partitions, so time.AddDate() seems to be a better fit than time.Add().

PR #23, opened 2019-06-04
Added changes to docker-compose so that we can run asset-inventory-ap…

aslape:
…i locally


README.md
```diff
@@ -85,6 +85,13 @@ building and testing a bit easier. The following actions are available through t
     Generate the project code documentation and make it viewable
     locally.
 
+To run locally:
+    1. In api.yml at x-runtime->stats->output (around line 9) change “DATADOG” to “NULL”
```
bkrebsbach: Question: Any reason not to just change these to `NULL` by default, similar to what you've done in the docker-compose config?

aslape: Honestly, I'm not sure. I decided not to push the api.yml changes because I didn't want to risk messing up anything relating to datadog in production.

README.md
```diff
@@ -85,6 +85,13 @@ building and testing a bit easier. The following actions are available through t
...
+    2. In api.yml at paths->/v1/cloud/change->post->x-transportd->enabled (around line 78) comment out: - “metrics”
+    3. In api.yml at paths->/v1/cloud/ip/{ipAddress}->get->x-transportd->enabled (around line 127) comment out: - “metrics”
```
zlozano: I think you'll need to comment it out on the /ops paths too. Or just get rid of it entirely since this isn't the configuration used in production

docker-compose.yaml
```diff
@@ -17,16 +19,36 @@ services:
       - SERVERFULL_RUNTIME_CONNSTATE_NEWCOUNTER=http.server.connstate.new
       - SERVERFULL_RUNTIME_LOGGER_OUTPUT=STDOUT
       - SERVERFULL_RUNTIME_LOGGER_LEVEL=INFO
-      - SERVERFULL_RUNTIME_STATS_OUTPUT=DATADOG
+      - SERVERFULL_RUNTIME_STATS_OUTPUT=NULL
       - SERVERFULL_RUNTIME_STATS_DATADOG_PACKETSIZE=32768
```
zlozano: If you set STATS output to NULL, you should be able to remove all of the `SERVERFULL_RUNTIME_STATS_DATADOG_` settings

aslape: Ok, cool. Will do!

README.md
bkrebsbach: That makes sense. The api.yaml in this service is meant to be a functional example, not necessarily what we would run in production. It seems worth it, to me, to just drop DATADOG/metrics from api.yaml for the sake of making the compose environment buildable by default.

aslape: Ah, okay. Didn't know that. I'll go ahead and remove those then. Thanks!

zlozano: Now that these items are already removed, this "run locally" sections can simply read: execute `make run`, or something to that effect.

aslape: Oh yea... Silly mistake. Thanks for catching that.

PR #24, opened 2019-08-07
Add ability to delete by partition name

kconwayatlassian:


zlozano:
question: what was the rationale behind replacing the existing DELETE endpoint vs. creating a new one?

FWIW the intention behind the current delete was to easily archive partitions older than a certain date. Without this, the client would have to fetch all partitions, perform the logic of which partitions need to be deleted, and issue one ore more delete requests for specific partitions which met their criteria

kconwayatlassian:
> question: what was the rationale behind replacing the existing DELETE endpoint vs. creating a new one?
> 
> FWIW the intention behind the current delete was to easily archive partitions older than a certain date. Without this, the client would have to fetch all partitions, perform the logic of which partitions need to be deleted, and issue one ore more delete requests for specific partitions which met their criteria

The issue with the current DELETE is that our original use case for it has been somewhat invalidated. Dropping old partitions is only valid if we can guarantee that all records in the partition have been replaced by an updated record in a newer partition. Long-lived assets require that we preserve old partitions until that asset is removed or updated. Deleting all partitions older than X amount of time is very likely to lead to data loss.

api.yaml
```diff
@@ -302,11 +298,9 @@ paths:
           async: false
           request: >
             {
-              #! if .Request.Query.days !#
-              "days": #!index .Request.Query.days 0!#
-              #! end !#
+              "name": #!index .Request.Query.name 0!#
             }
-          success: '{"status": 200, "bodyPassthrough": true}'
+          success: '{"status": 204, "bodyPassthrough": true}'
```
bkrebsbach: Nit: Given that this returns a 204, you should probably drop the `bodyPassthrough`

api.yaml
```diff
@@ -302,11 +298,9 @@ paths:
...
           error: '{"status": 500, "bodyPassthrough": true}'
```
LorenBoren: Question: I think it would be useful to let the user know if the partition they specified does not exist by returning a 400 with a response body saying that the partition doesn't exist?

kconwayatlassian: Question (@LorenBoren ): The current implementation returns 204 for all calls to delete. If you were the one calling operating the service and manually deleting partitions, how would you handle the 400 or 404 case differently than the 204?

LorenBoren: 204 indicated success. If I call the endpoint with a partition name that doesn't exist, but receive a 204, then I assume that the partition existed and was deleted, but that's not what happened. It's true that after the call I make there's no partition, so the end result is the same, but if I supplied a partition name that doesn't exist, then that means there's something wrong on my end that should be fixed, whether it's my understanding of the application, the data I'm looking at telling me what partitions exist, or a simple typo.

kconwayatlassian: @LorenBoren I think it's a good point. I was considering the retry case where the same DELETE is executed multiple times and I intended all of the same calls to return 204 rather than a scenario like:
- Execute DELETE
- I lose my connection while waiting for the response but the table was dropped successfully
- Execute DELETE -> 404

However, this API is pretty well exclusively for operators as it's unclear how and when use of it would be automated. In that case, I think having the additional context in the response would be useful. I'll open a ticket to address this. Thanks for pointing this out.

PR #26, opened 2019-08-14
added record count to partitions

sydneyteh96:


Dockerfile
```diff
@@ -3,9 +3,9 @@ RUN mkdir -p /go/src/github.com/asecurityteam/asset-inventory-api
 WORKDIR $GOPATH/src/github.com/asecurityteam/asset-inventory-api
 COPY --chown=sdcli:sdcli . .
 RUN sdcli go dep
-RUN go get -u github.com/gobuffalo/packr/v2/packr2
+RUN GO111MODULE=on go get github.com/gobuffalo/packr/v2/packr2
```
bkrebsbach: Task: Nice find! I believe you'll also need to add `GOFLAGS='-mod=vendor'` so go uses the vendor directory created by the `sdcli go dep` stage. See: https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

pkg/storage/db.go
```diff
@@ -9,8 +9,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/asecurityteam/asset-inventory-api/pkg/domain"
-	_ "github.com/lib/pq" // must remain here for sql lib to find the postgres driver
+	"github.com/asecurityteam/asset-inventory-api/pkg/domain" // must remain here for sql lib to find the postgres driver
```
bkrebsbach: Nit: comment appears to be duplicated from the pq package below

pkg/storage/db.go
```diff
@@ -295,14 +295,16 @@ func handleRollback(tx *sql.Tx, err error) error {
 	return err
 }
 
-// GetPartitions fetches the created partitions in the database
+// GetPartitions fetches the created partitions and gets each record count in the database
 func (db *DB) GetPartitions(ctx context.Context) ([]domain.Partition, error) {
-	stmt := "SELECT name, created_at, partition_begin, partition_end FROM partitions ORDER BY partition_end DESC"
+	//stmt := `SELECT name, created_at, partition_begin, partition_end
```
bkrebsbach: Nit: These comments lines should be deleted.

scripts/2_create.sql
```diff
@@ -65,7 +66,8 @@ IF NOT EXISTS partitions
     name VARCHAR PRIMARY KEY,
     created_at TIMESTAMP NOT NULL,
     partition_begin DATE NOT NULL,
-    partition_end DATE NOT NULL
+    partition_end DATE NOT NULL,
+    total_records INT
```
bkrebsbach: Question: it looks like you're adding this total_records value to the table, but it's not being used. Can this be removed?

pkg/storage/db.go
```diff
@@ -320,11 +320,17 @@ func (db *DB) GetPartitions(ctx context.Context) ([]domain.Partition, error) {
 			End:       end,
 		})
 	}
-
 	if err := rows.Close(); err != nil {
 		return nil, err
 	}
-
+	for _, v := range partitions {
+		row := db.sqldb.QueryRow("SELECT count(*) FROM " + v.Name)
```
kconwayatlassian: task: This needs to use the [`QueryRowContext`](https://golang.org/pkg/database/sql/#Conn.QueryRowContext) method similar to how the other query above uses `QueryContext`. Passing the context variable around makes sure that our system behaves correctly if we decide to stop doing work. For example, if someone makes the API call to list all partitions but then closes their connection half-way through processing then we don't want to continue making queries to the database. The context is how we signal that a request is cancelled and work should stop.

pkg/storage/db.go
```diff
@@ -320,11 +320,17 @@ func (db *DB) GetPartitions(ctx context.Context) ([]domain.Partition, error) {
...
+		var count int
+		if err := row.Scan(&count); err != nil {
+			return nil, err
+		}
+		v.Count = count
```
kconwayatlassian: task: Unfortunately, this doesn't work the way you expect it to in Go. I put together a playground to demonstrate why: https://play.golang.org/p/xk9C3ARk3B1. There is also a good write-up about this at https://dave.cheney.net/2017/04/29/there-is-no-pass-by-reference-in-go.

tl;dr You need adjust the loop to look like:

```golang
for offset := range partitions {
  // <query for the count>
  partitions[offset].Count = count
}
```

I recommend writing or modifying one of the unit tests in db_test.go in order to verify that the we get the correct behavior from setting the counts.

scripts/2_create.sql
sydneyteh96: Removed. Will appear in my next commit. Thank you!

pkg/storage/db.go
sydneyteh96: Fixed that. Thank you

pkg/storage/db.go
sydneyteh96: Changed to QueryRowContext. Thank you

pkg/storage/db.go
sydneyteh96: Will delete that. Thank you

pkg/storage/db.go
zlozano: suggestion: as the number of partitions increase over time, we will pay a cost of additional round trips. You don't need to make this change at the moment, but it may be worth getting a baseline for the performance of using  multiple queries, and comparing it agains building up one query to execute: https://stackoverflow.com/a/10185612

Dockerfile
sydneyteh96: Ok. Added the parameter. Thank you for your opinion.
 @zlozano

pkg/storage/db.go
sydneyteh96: Thank you for telling

zlozano:
task: can you add/augment an integration test for this addition?

PR #28, opened 2019-08-21
After build failed, I removed GOFLAGS='-mod=vendor', so that go get c…

sydneyteh96:
…an run to build binary files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ COPY --chown=sdcli:sdcli . .
 RUN sdcli go dep
 RUN GO111MODULE=on go get github.com/gobuffalo/packr/v2/packr2
 RUN packr2
-RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -o /opt/app main.go
+RUN CGO_ENABLED=0 GOOS=linux GOFLAGS="-mod=vendor" GO111MODULE=on go build -a -o /opt/app main.go
 
 ##################################
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ RUN mkdir -p /go/src/github.com/asecurityteam/asset-inventory-api
 WORKDIR $GOPATH/src/github.com/asecurityteam/asset-inventory-api
 COPY --chown=sdcli:sdcli . .
 RUN sdcli go dep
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -3,7 +3,7 @@ RUN mkdir -p /go/src/github.com/asecurityteam/asset-inventory-api
 WORKDIR $GOPATH/src/github.com/asecurityteam/asset-inventory-api
 COPY --chown=sdcli:sdcli . .
 RUN sdcli go dep
-RUN GO111MODULE=on GOFLAGS='-mod=vendor' go get github.com/gobuffalo/packr/v2/packr2
```
bkrebsbach: Question: why was this breaking the build?

Dockerfile
```diff
@@ -3,7 +3,7 @@ RUN mkdir -p /go/src/github.com/asecurityteam/asset-inventory-api
...
+RUN GO111MODULE=on go get github.com/gobuffalo/packr/v2/packr2
 RUN packr2
 RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -o /opt/app main.go
```
bkrebsbach: Task: I'm not sure if the GOFLAGS var is needed for the `go get` on line 6, but it should be passed here on line 8 so `go build` uses the vendored dependencies, rather than pulling directly from source.

zlozano: +1

Dockerfile
zlozano: the mod=vendor flag disables `go get` entirely. We need to go get this package so we can execute the packr2 command.

bkrebsbach: TIL, thanks!

go.mod
```diff
@@ -29,5 +29,4 @@ require (
 	golang.org/x/net v0.0.0-20190509222800-a4d6f7feada5 // indirect
 	golang.org/x/sys v0.0.0-20190509141414-a5b02f93d862 // indirect
 	golang.org/x/tools v0.0.0-20190510151030-63859f3815cb // indirect
-	github.com/lib/pq v1.0.0
```
zlozano: :fistshake: gomod!

Looks like this is an issue that won't be fixed in go 1.12: https://github.com/golang/go/issues/29702

We will need to add this line back in manually since `go get -u` and/or `go mod tidy` seem to not recognize unused imports

PR #29, opened 2019-09-09
Bulk paginated api for backfilling

ensary:
Asking for early feedback.
To be updated with tests and possibly changes to queries once I look into performance.


main.go
```diff
@@ -38,6 +38,11 @@ func main() {
 		StatFn:  domain.StatFromContext,
 		Fetcher: dbStorage,
 	}
+	fetchAllByTime := &v1.CloudFetchAllByTimestampHandler{
```
gcase555: question: why not call this `fetchAssetsByTime` to match the servefull function name here https://github.com/asecurityteam/asset-inventory-api/pull/29/files#diff-7ddfb3e035b42cd70649cc33393fe32cR62

ensary: Does `fetchAllAssetsByTime` sound like a better name? (I am trying to convey this does bulk vs. finding a single thing)

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -108,6 +117,46 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+func (h *CloudFetchAllByTimestampHandler) Handle(ctx context.Context, input CloudAssetFetchAllByTimestampParameters) (CloudAssets, error) {
+	logger := h.LogFn(ctx)
+
+	ts, e := time.Parse(time.RFC3339Nano, input.Timestamp)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return CloudAssets{}, InvalidInput{Field: "time", Cause: e}
+	}
+
+	if input.Count == nil {
```
gcase555: question: could the `input.Count` be equal to 0? wouldn't this check miss it if the count was 0?

main.go
gcase555: Yes I like that name, especially using `assets` versus `asset` to distinguish its looking for multiple instead of a singular asset.

pkg/handlers/v1/cloud_fetch.go
ensary:
```    schema:
            type: "integer"
            minimum: 1
```
the API specs would not allow zero

main.go
ensary: Changing.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -38,6 +39,14 @@ type CloudAssetFetchByHostnameParameters struct {
 	Timestamp string `json:"time"`
 }
 
+// CloudAssetFetchAllByTimestampParameters represents the incoming payload for bulk fetching cloud assets for point in time with optional pagination
+type CloudAssetFetchAllByTimestampParameters struct {
+	Timestamp string `json:"time"`
+	// we use the pointer type to detect if the value was not present in input as otherwise the int variable would be 0, which is a valid input
```
zlozano: question: if 0 is invalid, why not just check to see if value == 0? Also, it looks like you are using 0 for the default value of offset. Is it a pointer here for consistency?

ensary: The pointer use if for "generic" way to distinguish between the case when the value passed was zero vs. no value passed. I think I can change this to integer and just treat 0 as 'no value passed or 0 passed'. Good point, it makes little sense in this particular case.

pkg/storage/db.go
```diff
@@ -60,6 +60,35 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
 	"    aws_resources ON " +
 	"        latest.aws_resources_id = aws_resources.id;"
 
+// This query is used to retrieve all the 'active' resources (i.e. those with assigned IP/Hostname) for specific date
+// TODO - run performance analysis, possibly do something else on SQL level (optimize query or adjust schema)
```
zlozano: +1

zlozano:
LGTM so far

gcase555:
lgtm so far as well

api.yaml
```diff
@@ -85,6 +85,75 @@ paths:
...
+          request: >
+            {
+              "time": "#!index .Request.Query.time 0!#",
+              "count": #!index .Request.Query.count 0!# #! if .Request.Query.offset !# ,
+              "offset": #!index .Request.Query.offset 0!#
+              #! end !#
+            }
+          success: '{"status": 200, "bodyPassthrough": true}'
+          error: >
+            {
```
bkrebsbach: Suggestion: After playing around with this syntax a bunch, I personally find something like this to be a little easier to read, since it separates the JSON response from the template syntax. There's more duplication here though, so I have no strong feelings either way.

```json
            #! if eq .Response.Body.errorType "NotFound" !#
            { "status": 404 }
            #! else if eq .Response.Body.errorType "DependencyFailure" !#
            { "status": 502, "body": { "code": 502, "status": "Bad Gateway", "reason": "#!.Response.Body.errorMessage!#" }}
            #! else !#
            { "status": 500, "body": { "code": 500, "status": "Internal Server Error", "reason": "#!.Response.Body.errorMessage!#" }}
            #! end !#
```

api.yaml
```diff
@@ -85,6 +85,75 @@ paths:
...
+          required: true
+          schema:
+            type: "integer"
+            minimum: 1
+        - name: "offset"
+          in: "query"
+          description: "Offset of the first cloud asset to return for paged API access. Zero-based."
+          required: false
+          schema:
+            type: "integer"
```
kconwayatlassian: suggestion: A numeric offset exposes the use of a SQL datastore as an implementation detail. If there's another marker we could use, such as the timestamp of the last element we returned, then it would help the API be agnostic of the datastore.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -108,6 +117,46 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+func (h *CloudFetchAllAssetsByTimeHandler) Handle(ctx context.Context, input CloudAssetFetchAllByTimestampParameters) (CloudAssets, error) {
+	logger := h.LogFn(ctx)
+
+	ts, e := time.Parse(time.RFC3339Nano, input.Timestamp)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return CloudAssets{}, InvalidInput{Field: "time", Cause: e}
+	}
+
+	if input.Count == nil {
```
kconwayatlassian: suggestion: @gcase555 comments here. We could also allow a zero input and replace it with a default value.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -108,6 +117,46 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+		e = errors.New("missing or malformed required parameter count")
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return CloudAssets{}, InvalidInput{Field: "count", Cause: e}
+	}
+
+	var offset uint
+	if input.Offset != nil {
+		offset = *input.Offset
```
kconwayatlassian: suggestion: There is a difference between an optional value in the JSON payload and pointer values on the struct. We usually only introduce pointer values when we need to distinguish between explicit zero and implicit zero. My understanding is that it does not matter for offset because implicit and explicit zero have the same meaning. I recommend, for this case, having the optional input value described in the JSON schema but having a non-pointer type field for the struct.

api.yaml
ensary: ACK. Good idea!

pkg/handlers/v1/cloud_fetch.go
ensary: Got it.

zlozano:
Approving this in the WIP state. Still would like integration tests.

api.yaml
```diff
@@ -138,8 +138,8 @@ paths:
           request: >
             {
               "time": "#!index .Request.Query.time 0!#",
-              "count": #!if .Request.Query.count !# #!index .Request.Query.count 0!# #! else !# 100 #! end !# #!if .Request.Query.type !# ,
-              "type": "#!index .Request.Query.type 0!#" #! end !#
+              "count": #!if .Request.Query.count !# #!index .Request.Query.count 0!# #! else !# 100 #! end !# ,
```
zlozano: this is not really a suggestion, mostly posing the question: would this default value be better if embedded as a value inside the app? If in the app, it can be used immediately by anyone who uses this service. If left here in the api descriptor, it must be duplicated by anyone who uses it.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -42,9 +56,42 @@ type CloudAssetFetchByHostnameParameters struct {
 // CloudAssetFetchAllByTimestampParameters represents the incoming payload for bulk fetching cloud assets for point in time with optional pagination
 type CloudAssetFetchAllByTimestampParameters struct {
 	Timestamp string `json:"time"`
-	// we use the pointer type to detect if the value was not present in input as otherwise the int variable would be 0, which is a valid input
-	Count  *uint `json:"count"`
-	Offset *uint `json:"offset"`
+	Count     uint   `json:"count"`
+	Offset    uint   `json:"offset"`
```
zlozano: task: AFAICT this isn't used anymore

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -117,44 +164,113 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+	logger := h.LogFn(ctx)
+	params, e := fetchAllByTimeStampParametersForToken(input.PageToken)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return PagedCloudAssets{}, InvalidInput{Field: "PageToken", Cause: e}
+	}
+	ts, e := time.Parse(time.RFC3339Nano, params.Timestamp)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return PagedCloudAssets{}, InvalidInput{Field: "PageToken", Cause: e}
```
zlozano: suggestion: these error messages leak out details of our `nextPageToken`. Maybe we can return something more generic e.g. "Invalida page token provided" while logging the details of the error.

pkg/storage/db.go
```diff
@@ -546,48 +545,49 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
 		var timestamp time.Time
 
 		err = rows.Scan(&row.ARN, &ipAddress, &hostname, &isPublic, &isJoin, &timestamp, &row.AccountID, &row.Region, &row.ResourceType, &metaBytes)
+		if err != nil {
```
zlozano: +1 thanks!

pkg/handlers/v1/cloud_fetch.go
ensary: Changing

api.yaml
ensary: Do we have a good example? It looks like a good idea to me.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -42,9 +56,42 @@ type CloudAssetFetchByHostnameParameters struct {
...
 	Timestamp string `json:"time"`
-	// we use the pointer type to detect if the value was not present in input as otherwise the int variable would be 0, which is a valid input
-	Count  *uint `json:"count"`
-	Offset *uint `json:"offset"`
+	Count     uint   `json:"count"`
+	Offset    uint   `json:"offset"`
+	Type      string `json:"type"`
+}
+
+func (p *CloudAssetFetchAllByTimestampParameters) toNextPageToken() (string, error) {
```
kconwayatlassian: nit: Personally, I prefer not having behaviors/methods attached to data containers. My recommendation is generally that 1) interfaces define behaviors but do not contain data, 2) structs contain data but do not implement behaviors, and 3) a struct may implement and interface but then it must not contain data. The phrasing of "contains data" here means that structs attributes are exported and we expect people to use them directly. The goal in that philosophy is to ensure that we don't fall into traditional class patterns where we're trying to model state and ways of manipulating that state. It's not perfect and there are no "hard lines" in practice but I find it useful.

For example, instead of `func(p *CloudFetchAllbyTimestampParameters) toNextPageToken()` I'd suggest `func NextPageToken(p CloudFetchAllByTimestampParameters)`.  Practically speaking, these are both the same. The receiver pointer notation like `func(*Something)` is intended to replace old C-style struct methods that eventually became class/instance methods. This is really more of a philosophical change where receiver functions operate on the state of the instance they are attached to and the non-receiver style is a pure function that operates on a given input.

Again, they aren't that different in most cases. I just find it more comfortable to avoid any accidental state that _might_ show up in class-style methods.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -117,44 +164,113 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
 	}
 	if len(assets) == 0 {
-		return CloudAssets{}, NotFound{ID: "any"}
+		return PagedCloudAssets{}, NotFound{ID: "any"}
 	}
 
-	return extractOutput(assets), nil
+	nextPageToken, e := input.toNextPageToken()
+	if e != nil {
+		logger.Error(logs.StorageError{Reason: e.Error()})
```
kconwayatlassian: question: Why do we continue here instead of returning the storage error?

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -117,44 +164,113 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+	var offset uint // do not offset as this is the first page
+	assets, e := h.Fetcher.FetchAll(ctx, ts, input.Count, offset, assetType)
 	if e != nil {
 		logger.Error(logs.StorageError{Reason: e.Error()})
-		return CloudAssets{}, e
+		return PagedCloudAssets{}, e
 	}
 	if len(assets) == 0 {
-		return CloudAssets{}, NotFound{ID: "any"}
+		return PagedCloudAssets{}, NotFound{ID: "any"}
```
kconwayatlassian: suggestion: This was not introduced in this PR, but we would usually keep custom error types in the `domain` package since they are used to communicate across interface boundaries.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -117,44 +164,113 @@ func (h *CloudFetchByHostnameHandler) Handle(ctx context.Context, input CloudAss
...
+		return PagedCloudAssets{}, InvalidInput{Field: "pageToken", Cause: e}
+	}
+	ts, e := time.Parse(time.RFC3339Nano, params.Timestamp)
+	if e != nil {
+		logger.Info(logs.InvalidInput{Reason: e.Error()})
+		return PagedCloudAssets{}, InvalidInput{Field: "pageToken", Cause: e}
+	}
+
+	//generic error to report to caller to avoid exposing the internal token structure NB, the specific error is still logged
+	tokenError := errors.New("malformed pageToken")
```
kconwayatlassian: nit: This could be a package constant/variable and re-used rather than redefined on each call.

pkg/storage/db.go
```diff
@@ -546,48 +545,49 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
...
-				if isPublic {
-					tempMap[row.ARN].PublicIPAddresses = newArray
-				} else {
-					tempMap[row.ARN].PrivateIPAddresses = newArray
-				}
+				tempMap[row.ARN].Hostnames = append(tempMap[row.ARN].Hostnames, hostname.String)
+			}
+		}
+		found = false
+		var ipAddresses *[]string
```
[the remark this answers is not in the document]
ensary: It's outside this PR (the change is indentation) but I will look into fixing this in the next one.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -42,9 +56,42 @@ type CloudAssetFetchByHostnameParameters struct {
...
+		Timestamp: p.Timestamp,
+		Count:     p.Count,
+		Offset:    p.Offset + p.Count,
+		Type:      p.Type,
+	}
+	js, err := json.Marshal(nextPageParameters)
+	if err != nil {
+		return "", err
+	}
+	token := base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(js)
```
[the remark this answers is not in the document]
ensary: Intentional. Base32 is case insensitive, which makes use in path (or part of URL) safer. Not exactly a big deal in our case, but the thinking was to make this as portable as possible.

pkg/handlers/v1/cloud_fetch.go
ensary: Excellent explanation, esp the "receiver functions operate on the state of the instance they are attached to and the non-receiver style is a pure function that operates on a given input" part.

pkg/handlers/v1/cloud_fetch.go
ensary: The current page of results is valid, and the case where we can not generate the token from valid parameters is unlikely (if at all possible), but I did not want to ignore the possible error condition that we did not foresee. So the logic currently is - return valid results (that's what client requested), log any issue with generating token for the next page, and let the next request fail.
One clear action item - I should not use StorageError for this, as it has nothing to do with storage, but I am open to failing loudly if we can't  generate token for the next page even when we have valid results for the current one.

pkg/handlers/v1/cloud_fetch.go
ensary: Great point

api.yaml
zlozano: Not that I know of. One approach would be to have an attribute inside your handler for a default page size which is populated from an environment variable.

PR #30, opened 2019-09-17
added 404 route for deleting partitions that do not exist

dscgarcia:
per SID-43, DELETE endpoint can now return a 404 if the partition requested to be deleted does not exist, returns new "NotFoundPartition" error


zlozano:
It appears the lint step failed. You should be able to run `make lint` locally to verify the issue.

tests/storage_test.go
```diff
@@ -522,6 +522,46 @@ func TestDeletePartitions(t *testing.T) {
...
+	assert.True(t, ok)
+
+	partitions, err = dbStorage.GetPartitions(context.Background())
+	require.NoError(t, err)
+	partitionNames = make([]string, 0, len(partitions))
+	for _, part := range partitions {
+		partitionNames = append(partitionNames, part.Name)
+	}
+
+	assert.Equal(t, numPartitions+1, len(partitions))
```
kconwayatlassian: task: I believe you meant this to be `len(partitionNames)`. The patch is failing linting because the above `partitionNames` variable is populated but then never used.

tests/storage_test.go
```diff
@@ -522,6 +522,46 @@ func TestDeletePartitions(t *testing.T) {
...
+	err = dbStorage.DeletePartitions(context.Background(), nonexistentPartition)
+	require.Error(t, err)
+	_, ok := err.(domain.NotFoundPartition)
+	assert.True(t, ok)
+
+	partitions, err = dbStorage.GetPartitions(context.Background())
+	require.NoError(t, err)
+	partitionNames = make([]string, 0, len(partitions))
+	for _, part := range partitions {
+		partitionNames = append(partitionNames, part.Name)
```
zlozano: The linter is complaining that `partitionNames` is never used. You can run this locally by running `make lint`

PR #33, opened 2019-12-12
Switch db reads to replica

ensary:
This is the first pass on using read replica for lookups, I picked up the branch @dscgarcia started work on and introduced some minor changes.
The environment variable to configure PSQL are now under AIAPI_ prefix to avoid ambiguity and follow the naming convention we use in other services.
NB, there likely to be more changes once I test by deploying to dev with read replica configured. These will come as the separate PR together with more tests.


docker-compose.it.yml
```diff
@@ -8,6 +8,11 @@ services:
       - "POSTGRES_DATABASENAME=assetmgmt"
```
gcase555: reviewers note:

it appears this docker-compose file is used for integration testing https://github.com/asecurityteam/asset-inventory-api/blob/69f69be36474dc3e9415d57859c8f9a45b2c18cf/Makefile#L25

docker-compose.yaml
```diff
@@ -22,11 +22,16 @@ services:
       - SERVERFULL_RUNTIME_STATS_OUTPUT=NULL
       - SERVERFULL_RUNTIME_SIGNALS_INSTALLED=OS
       - SERVERFULL_RUNTIME_SIGNALS_OS_SIGNALS=15 2
-      - "POSTGRES_PASSWORD=password"
-      - "POSTGRES_USERNAME=user"
-      - "POSTGRES_DATABASENAME=assetmgmt"
-      - "POSTGRES_HOSTNAME=postgres"
-      - "POSTGRES_PORT=5432"
+      - "AIAPI_POSTGRES_PASSWORD=password"
```
gcase555: question:
what is the motivation to add the prefix `AIAPI_` to these `POSTGRES` environment variables? Why is it also not needed above in the `docker-compose.it.yml`?

docker-compose.it.yml
ensary: Yes, it is

docker-compose.yaml
ensary: Because docker-compose.it.yml uses the code that does not include top-level component for configuration and reads env vars directly (see code in tests/*). I am not  100% confident this is the right thing to do, but that's the way it is now.

ensary: Not 100% confident - not re: the service, but re: the tests. The prefix is a well established pattern in the rest of our services.

pkg/storage/db.go
```diff
@@ -117,16 +117,11 @@ func (db *DB) RunScript(ctx context.Context, name string) error {
 }
 
 // Init initializes a connection to a Postgres database according to the environment variables POSTGRES_HOST, POSTGRES_PORT, POSTGRES_USER, POSTGRES_PASSWORD, POSTGRES_DATABASE
-func (db *DB) Init(ctx context.Context, postgresConfig *PostgresConfig) error {
+func (db *DB) Init(ctx context.Context, host string, port string, user string, password string, dbname string, partitionTTL int) error {
```
gcase555: nit:

This is borderline crossing into too many parameter territory at 6 but I see it reduces the the redundant variables we must set below.

I like it in this case but if we need to add more variables to this `Init` in the future then we might consider going back to the struct.

pkg/storage/read_config.go
```diff
@@ -0,0 +1,49 @@
+package storage
+
+import (
+	"context"
+)
+
+// PostgresReadConfig contains the Postgres database configuration arguments for the ReadReplica
+type PostgresReadConfig struct {
+	Hostname     string
+	Port         string
```
gcase555: question: why is port a string?

scripts/0_wipe.sql
```diff
@@ -1 +1 @@
-DROP DATABASE %s;
\ No newline at end of file
+DROP DATABASE IF EXISTS %s;
```
gcase555: 👌 
nit: no newline at end of file

pkg/storage/read_config.go
ensary: This is the same with existing PostgresConfig. 
I will try to change it to UInt16 next time I touch it and see what breaks.

ensary: Done

PR #34, opened 2020-01-02
Switch db reads to replica

ensary:
Prevent replica DB connection initialization from attempting to create database and manage schema.
This is the final pass re: read replica support. The Docker image with changes from this branch is functional and correctly directs reads to replica.
NB - the schema management logic is going to change once we have tooling in place.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -9,9 +9,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/asecurityteam/asset-inventory-api/pkg/domain"
 	_ "github.com/lib/pq" // must remain here for sql lib to find the postgres driver
 	"github.com/pkg/errors"
+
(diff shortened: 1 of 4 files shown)


pkg/storage/db.go
```diff
@@ -9,9 +9,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/asecurityteam/asset-inventory-api/pkg/domain"
 	_ "github.com/lib/pq" // must remain here for sql lib to find the postgres driver
 	"github.com/pkg/errors"
+
+	"github.com/asecurityteam/asset-inventory-api/pkg/domain"
```
ensary: Interesting. It seems like `goimports -local github.com/asecurityteam` now puts this on a separate section with a blank line before it. Does anyone have any comments/insight re: this?

gcase555: Possibly because this is an internal project module versus a public one?

PR #36, opened 2020-01-14
Normalized schema

ensary:
The proposed new schema (with manual migration and performance testing so far, automation for both is to be done in scope of different PRs) was tested for performance comparison with existing one using the queries in this PR and production-like data. 
The storage was populated (for both cases) with ~10 mln resources and ~45 mln allocation/de-allocation events.
The query execution time for look-ups  (both hits and misses) using new schema did not exceed 100ms, which brings us well within reasonable execution time per request.

Caveats:
* not ALL the queries using old schema resulted in multi-second look-ups, but some  took 5+ seconds to complete with the data set I used. 
* the new schema was specifically compared against worst performing cases for current schema, and is more predictable performance wise (i.e. varies less)
* the partitioning is no longer used, as there is no way to effectively partition on time (the open date of the event can be at any point in the past, so we would have to potentially keep partitions forever). In addition, based on the numbers from performance tests with proposed schema changes the additional complexity of maintaining partitions is not justified as the performance is good enough as-is

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -76,10 +76,10 @@ from aws_private_ip_assignment ia
 where ia.private_ip = $1
   and ia.not_before < $2
   and (ia.not_after is null or ia.not_after > $2)
-`
+` // nolint: varcheck
 
 // Query to find resource by public IP using v2 schema
-const resourceByPublicIpQuery = `select ia.public_ip,
+const resourceByPublicIPQuery = `select ia.public_ip,
        res.arn_id,
        res.meta,
        ar.region,
@@ -93,7 +93,7 @@ from aws_public_ip_assignment ia
 where ia.public_ip = $1
   and ia.not_before < $2
   and (ia.not_after is null or ia.not_after > $2)
-`
+` // nolint: varcheck
 
 // Query to find resource by hostname using v2 schema
 const resourceByHostnameQuery = `select ia.aws_hostname,
@@ -110,7 +110,7 @@ from aws_public_ip_assignment ia
 where ia.aws_hostname = $1
   and ia.not_before < $2
   and (ia.not_after is null or ia.not_after > $2)
-`
+` // nolint: varcheck
 
 // This query is used to retrieve all the 'active' resources (i.e. those with assigned IP/Hostname) for specific date
 const bulkResourcesQuery = `
@@ -139,7 +139,7 @@ LIMIT $3
 OFFSET $4
 `
 
-//TODO Optimized query to retrieve all the 'active' resources utilising v2 schema. Out of scope currently.
+//TODO Optimized query to retrieve all the 'active' resources utilizing v2 schema. Out of scope currently.
 
 // DB represents a convenient database abstraction layer
 type DB struct {
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -61,8 +61,58 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
 	"    aws_resources ON " +
 	"        latest.aws_resources_id = aws_resources.id;"
 
+// Query to find resource by private IP using v2 schema
+const resourceByPrivateIPQuery = `select ia.private_ip,
+       res.arn_id,
+       res.meta,
+       ar.region,
+       rt.resource_type,
+       aa.account
+from aws_private_ip_assignment ia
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -61,8 +61,61 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
 	"    aws_resources ON " +
 	"        latest.aws_resources_id = aws_resources.id;"
 
+// Query to find resource by private IP using v2 schema
+// nolint
```
gcase555: question: what linting rules in particular does this break? (just curious)

scripts/3_create_v2.sql
```diff
@@ -0,0 +1,57 @@
...
+(
+    id              bigserial primary key,
+    not_before      timestamp not null,
+    not_after       timestamp,
+    private_ip      inet,
+    aws_resource_id int       not null,
+    foreign key (aws_resource_id) references aws_resource (id)
+);
+
+create index idx_private_ip on aws_private_ip_assignment (private_ip);
```
gcase555: nit: missing newline

pkg/storage/db.go
```diff
@@ -61,8 +61,61 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
+const resourceByPrivateIPQuery = `select ia.private_ip,
```
gcase555: these queries are pretty straight forward joins with some time ranges in the where classes that make sense to me 👍

scripts/3_create_v2.sql
```diff
@@ -0,0 +1,57 @@
...
+
+create table aws_resource_type
+(
+    id            serial primary key,
+    resource_type varchar not null unique
+);
+
+create table aws_resource
+(
+    id                   bigserial primary key,
```
gcase555: use of `bigserial` here ✅

scripts/3_create_v2.sql
```diff
@@ -0,0 +1,57 @@
...
+    arn_id               varchar not null, /* nb, this is NOT full arn, it is resource ID like i-5325235 as ARN is redundant and can be reconstructed */
+    aws_account_id       int     not null, /* the account, type and region are immutable, so the lookup tables are referenced as the part of resource, not event */
+    foreign key (aws_account_id) references aws_account (id),
+    aws_region_id        int     not null,
+    foreign key (aws_region_id) references aws_region (id),
+    aws_resource_type_id int     not null,
+    foreign key (aws_resource_type_id) references aws_resource_type (id),
+    meta                 JSONB /* this is not optimal and is not correct as tags can change and changes need tracking */
```
gcase555: question: what lead you to choose JSONB here? what do you think could be more optimal in the future? (or does that need further investigation)

ensary: Nothing. This part is unchanged from current schema. We will refactor it later, I just wanted to limit the scope of changes, and we do not need to normalize tags for M1.

pkg/storage/db.go
ensary: Unused constant.

scripts/3_create_v2.sql
ensary: fixed, tnx

scripts/3_create_v2.sql
ensary: re: optimal in the future - we need to track the changes to the specific tags, possibly in  separate tables as they are mutable and should have their own not_before and not_after values. At least those that establish ownership and tell which CF stack or micros service the resource is part of are of interest. So we will at some point need to stop treating the tags as immutable attribute of AWS resource. Possibly in m2 (depending on if we will be able to do attribution based on "blob" tags fast enough or not).

scripts/3_create_v2.sql
```diff
@@ -0,0 +1,59 @@
...
+create index idx_public_ip on aws_public_ip_assignment (public_ip);
+
+create index idx_aws_hostname on aws_public_ip_assignment (aws_hostname);
+
+create table aws_private_ip_assignment
+(
+    id              bigserial primary key,
+    not_before      timestamp not null,
+    not_after       timestamp,
+    private_ip      inet,
```
dscgarcia: is there a reason private_ip is nullable? is there an occasion where we will do an insert into this table without a private IP address?

ensary: Nope, and this is a great catch. Fixing.

PR #37, opened 2020-01-30
Schema management for PSQL

ensary:
* All the table creation/deletion via [migrate](https://github.com/golang-migrate/migrate) that was chosen as the result of schema management tools comparison
* Both 'current' and 'new' schemas are supported (versions 1 and 2). During migration to 'new' current is left in place to support dual writes
* Minimum schema version enforcement on start via configuration setting
* Integration tests use migrate vs. touching the DB directly
* Getting schema version and going one step up or down are all exposed as API end-points to allow to un-tie schema changes from deployments 
* Logic to create the database if it does not exist is removed completely, as there is no known use-case, and we should not run the app with privileges that allow to do this in the first place

Items to be addressed in future PR(s):
* Integration test runs for all supported schema versions (i.e. "make sure it works with any schema version higher than X")
* Done : Additional unit tests around migrate (esp in storage layer). Current combination of unit+integration gives enough confidence to continue, but CAN be better


api.yaml
```diff
@@ -322,6 +322,76 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
+
+  /ops/pgsql/v1/schema/version/up:
+    get:
+      summary: "Migrate database schema one version up"
```
dscgarcia: awesome that we are exposing these (up/down) 
however, I did notice in the documentation for migrate that the Up() will migrate "applying all up migrations." If we make sure not to merge any migration scripts into the code base that we aren't ready to run then I think this is fine. But, similarly, the Down() will migrate "applying all down migrations" and I'm not sure we will want to go back that far if we find one of the "up" migrations unsuccessful. There are other methods that we could expose (in addition to these or instead of) to run the migrations (Steps() to indicate relative up/down migrations or Migrate() if we want to feed it a specific version) and still maintain the de-coupling from app deployment

ensary: Fixing, thank you!

api.yaml
```diff
@@ -322,6 +322,76 @@ paths:
...
+      responses:
+        200:
+          description: "Resulting Schema Version"
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SchemaVersion'
+      x-transportd:
+        backend: app
+        enabled:
```
gcase555: question: are there authorization plugins or ASAP like plugins `validateheaders` we should be enabling here or would that best be done in bitbucket?

db-migrations/000001_legacy_schema.down.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+DROP TABLE IF EXISTS aws_events_ips_hostnames;
+DROP TABLE IF EXISTS aws_resources;
+DROP TABLE IF EXISTS aws_ips;
+DROP TABLE IF EXISTS aws_hostnames;
+DROP TABLE IF EXISTS partitions;
+COMMIT;
```
gcase555: nit: missing newline at EOF

db-migrations/000002_m1_schema.down.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+-- NB the order is important because of foreign keys!
```
gcase555: ❤️the note, I ran into this when I was doing cleanups before

api.yaml
ensary: None of the auth logic is present in Github as it is specific to our deployment.

db-migrations/000001_legacy_schema.down.sql
ensary: fixed

PR #38, opened 2020-02-04
Dual-write events to new and old schema based on present schema version

ensary:
~~NB, this will be updated to cover the new code path with tests properly, but otherwise looks complete. Considering that's a lot of SQL to review, throwing it in early.~~ Tests are there.

Based on the current schema version, after the existing code wrote the event to the existing schema, if the migration to add new tables was complete, the storeV2() is invoked to perform additional write to new schema.
The db inserts are written with possibility of out-of-band and out-of-order events, for example release event arriving before assign (e.g. dual writes and subsequent back-fill) in mind.
Special timestamp value is used to express 'unknown assignment date' to allow for catch-up update to record an assignment after the release event was already recorded.
Public IPs assignments always arrive with matching collection of hostnames (verified across all known data in production), so there is no separate recording of hostname assign/free events.
Out of scope - tracking tag/metadata changes.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -873,7 +873,8 @@ $$
         update aws_public_ip_assignment
         set not_after=$1
         where public_ip = $2
-          and aws_resource_id = (select id from aws_resource where arn_id = $3);
+          and aws_resource_id = (select id from aws_resource where arn_id = $3)
+          and aws_hostname = $4;
         if not found then
             insert into aws_public_ip_assignment
                 (not_before, not_after, public_ip, aws_resource_id, aws_hostname)
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -37,10 +37,12 @@ const (
 )
 
 const (
-	// MinimumSchemaVersion Lowest version of database schema current code is able to handle
-	MinimumSchemaVersion uint = 1
 	// EmptySchemaVersion Version of database schema that cleans the database completely. Use cautiously!
 	EmptySchemaVersion uint = 0
+	// MinimumSchemaVersion Lowest version of database schema current code is able to handle
+	MinimumSchemaVersion uint = 1
+	// DualWriteSchemaVersion Lowest version of database schema that supports dual-writes
+	DualWriteSchemaVersion uint = 2
 )
 
 // can't use Sprintf in a const, so...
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -473,8 +475,125 @@ func (db *DB) Store(ctx context.Context, cloudAssetChanges domain.CloudAssetChan
...
+	}
+	ver, err := db.GetSchemaVersion(ctx)
+	if err != nil || ver < DualWriteSchemaVersion {
+		return nil
+	}
+	return db.StoreV2(ctx, cloudAssetChanges)
+}
 
+// StoreV2 an implementation of the Storage interface that records to a database using new schema, to replace Store
+func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetChanges) error {
```
dscgarcia: per the function comment, do we need to add this method to the Storage interface?

ensary: It's supposed to replace v1 in-place once we completely move over to new schema.
The primary reason to make it public right now is to be able to test it separately from the existing one.

ensary: I probably should have reflected this in the comment :)

pkg/storage/db.go
```diff
@@ -473,8 +475,125 @@ func (db *DB) Store(ctx context.Context, cloudAssetChanges domain.CloudAssetChan
...
+	tx, err := db.sqldb.Begin()
+	if err != nil {
+		return err
+	}
+	if err = db.ensureResourceExists(ctx, cloudAssetChanges, tx); err == nil {
+		arnID := resIDFromARN(cloudAssetChanges.ARN)
+		for _, val := range cloudAssetChanges.Changes {
+			for _, ip := range val.PrivateIPAddresses {
+				if strings.EqualFold(added, val.ChangeType) {
```
dscgarcia: Is it possible for an event to be an "UPDATE" IP address, and would we treat that the same as a DELETE? Or does the data come across as a DELETE plus an ADD? I don't know how AWS Config structures their events re: changes to ip address...

ensary: The only two types we support are UPDATE and DELETE (basically the if is copy/paste from existing Store). E.g. if an event with anything else for changetype arrived, we'd reject it at the swagger descriptor level as https://github.com/asecurityteam/asset-inventory-api/blob/master/api.yaml#L550 specifically lists ADDED and DELETED as supported event types.

pkg/storage/db.go
```diff
@@ -669,3 +788,100 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
...
+	//this way we can find unbalanced events while avoiding nullable not_before
+	//which provides minimal support for out-of-order events
+	const releasePublicIPQuery = `
+do
+$$
+    begin
+        update aws_public_ip_assignment
+        set not_after=$1
+        where public_ip = $2
+          and aws_resource_id = (select id from aws_resource where arn_id = $3);
```
dscgarcia: does the update statement also need to match on hostname?

ensary: The ips and hostnames always "match", as amazon generates them. I was looking into the old db which has assign/release events recorded separately.
On the other hand, I do not see a reason to NOT match for clarity/cleanliness so, will update :)

ensary: Added

pkg/storage/db.go
```diff
@@ -473,8 +475,125 @@ func (db *DB) Store(ctx context.Context, cloudAssetChanges domain.CloudAssetChan
 		}
 		return err
 	}
-	return tx.Commit()
+	if err = tx.Commit(); err != nil {
+		return err
+	}
+	ver, err := db.GetSchemaVersion(ctx)
+	if err != nil || ver < DualWriteSchemaVersion {
+		return nil
```
gcase555: question: is it desirable to not return the error here if `err` isn't null?

gcase555: although I get in the case of `ver < DualWriteSchemaVersion` being true err would be nil so we couldn't return anything. I just want to clarify that it's desirable to not return the error here.?

ensary: the err is for getting the schema version, which is unrelated to the write to the old schema (which was successful at this point). 
The reason for error when getting schema version is - the database was rolled back to the state before we introduced versioning/migration ('ver 0'), which is a totally valid condition from "old" schema perspective (as we already wrote data to old schema) but does not allow us to do dual-write.

pkg/storage/db.go
```diff
@@ -473,8 +475,125 @@ func (db *DB) Store(ctx context.Context, cloudAssetChanges domain.CloudAssetChan
...
+	if err = tx.Commit(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (db *DB) ensureResourceExists(ctx context.Context, cloudAssetChanges domain.CloudAssetChanges, tx *sql.Tx) error {
+	// reading pre-requisite to understand this query - https://www.postgresql.org/docs/current/queries-with.html
+	const createResourceQuery string = `
+with sel as (
```
gcase555: nit: notably other SQL in the codebase uses a mix of CAPS and lowercase for queries e.g. https://github.com/asecurityteam/asset-inventory-api/pull/38/files#diff-106eb1177cdcbe545695a4fdec2e2a32L48 I don't think we have a SQL style guideline anywhere that dictates we can't do lowercase so I'm ok with lowercase sql but the lack of uniformity will be confusing when onboarding someone. Eventually I'd like us to choose all lowercase or mixed casing.

ensary: We should be left with "new" SQL only by the time we are done with  remaining schema change related work. 
I am open to both lower-case and uppercase too though.

pkg/storage/db.go
```diff
@@ -473,8 +475,125 @@ func (db *DB) Store(ctx context.Context, cloudAssetChanges domain.CloudAssetChan
...
+    select val.arn_id,
+           val.region,
+           aws_region.id        as aws_region_id,
```
gcase555: nit: for some reason this spacing feels odd to me, I'm more use to SQL like `aws_region.id as aws_region_id`. This is probably coming from `goimport` though so we probably can't avoid it for now.

ensary: This is intellij sub-language support with auto-formatting specific to PSQL.
Goimports does not "grok" sublanguages.
It is an interesting topic re: formatting SQL as we can't expect everyone to use Jetbrains tools...

gcase555: I see, it can stay but just an observation we should keep in mind is the formatting in the SQL statements is un-uniform throughout the repo. Something we can improve in the future maybe.

PR #40, opened 2020-02-06
Update fetch by ip

sydneyteh96:
Update existing FetchByIP() function, so it fetches from database base on schema version. The code has passed all checks.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -89,7 +89,7 @@ const resourceByPrivateIPQuery = "SELECT ia.private_ip, " +
 	"			aa.account " +
 	"	FROM aws_private_ip_assignment ia " +
 	"		LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
-	"		LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
+	"		LEFT JOIN aws_region ar ON res.aws_region_id = ar.id " +
 	"		LEFT JOIN aws_resource_type rt ON res.aws_resource_type_id = rt.id " +
 	"		LEFT JOIN aws_account aa ON res.aws_account_id = aa.id " +
 	"	WHERE ia.private_ip = $1 " +
@@ -108,7 +108,7 @@ const resourceByPublicIPQuery = "SELECT " +
 	"			aa.account " +
 	"	FROM aws_public_ip_assignment ia " +
 	"		LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
-	"		LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
+	"		LEFT JOIN aws_region ar ON res.aws_region_id = ar.id " +
 	"		LEFT JOIN aws_resource_type rt ON res.aws_resource_type_id = rt.id " +
 	"		LEFT JOIN aws_account aa ON res.aws_account_id = aa.id " +
 	"	WHERE ia.public_ip = $1 " +
@@ -125,7 +125,7 @@ const resourceByHostnameQuery = "SELECT ia.aws_hostname, " +
 	"       aa.account " +
 	"	FROM aws_public_ip_assignment ia " +
 	"         LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
-	"         LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
+	"         LEFT JOIN aws_region ar ON res.aws_region_id = ar.id " +
 	"         LEFT JOIN aws_resource_type rt ON res.aws_resource_type_id = rt.id " +
 	"         LEFT JOIN aws_account aa ON res.aws_account_id = aa.id " +
 	"	WHERE ia.aws_hostname = $1 " +
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -573,7 +573,7 @@ from sel
          left join ins_aws_region using (region)
          left join ins_aws_account using (account)
          left join ins_aws_resource_type using (resource_type)
-on conflict do nothing 
+on conflict do nothing
 `
 	tagsBytes, _ := json.Marshal(cloudAssetChanges.Tags) // an error here is not possible considering json.Marshal is taking a simple map or nil
 	if _, err := tx.ExecContext(ctx,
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
-`
+const resourceByPrivateIPQuery = "SELECT ia.private_ip, " +
+	"			res.arn_id, " +
+	"			res.meta, " +
+	"			ar.region, " +
+	"			rt.resource_type, " +
+	"			aa.account " +
+	"	FROM aws_private_ip_assignment ia " +
+	"		LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
+	"		LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
```
dscgarcia: I think we missed this in the last PR, this should be "LEFT JOIN aws_region ar ON res.aws_region_id = ar.id " +", right?

pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
+	"			ia.public_ip, " +
+	"			ia.hostname, " +
+	"			res.arn_id, " +
+	"			res.meta, " +
+	"			ar.region, " +
+	"			rt.resource_type, " +
+	"			aa.account " +
+	"	FROM aws_public_ip_assignment ia " +
+	"		LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
+	"		LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
```
dscgarcia: same here, res.**aws_region_id**

pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
-`
+const resourceByHostnameQuery = "SELECT ia.aws_hostname, " +
+	"       res.arn_id, " +
+	"       res.meta, " +
+	"       ar.region, " +
+	"       rt.resource_type, " +
+	"       aa.account " +
+	"	FROM aws_public_ip_assignment ia " +
+	"         LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id " +
+	"         LEFT JOIN aws_region ar ON res.aws_account_id = ar.id " +
```
dscgarcia: same here, res.**aws_region_id**

pkg/storage/db.go
```diff
@@ -689,22 +688,33 @@ func (db *DB) insertNetworkChangeEvent(ctx context.Context, timestamp time.Time,
 
 // FetchAll gets all the assets present at the specified time
 func (db *DB) FetchAll(ctx context.Context, when time.Time, count uint, offset uint, typeFilter string) ([]domain.CloudAssetDetails, error) {
-	return db.runQuery(ctx, bulkResourcesQuery, when, typeFilter, count, offset)
+	return db.runQuery(ctx, 1, bulkResourcesQuery, when, typeFilter, count, offset)
```
dscgarcia: is there a reason the version is hardcoded to 1?

pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
-from aws_private_ip_assignment ia
-         left join aws_resource res on ia.aws_resource_id = res.id
-         left join aws_region ar on res.aws_account_id = ar.id
-         left join aws_resource_type rt on res.aws_resource_type_id = rt.id
-         left join aws_account aa on res.aws_account_id = aa.id
-where ia.private_ip = $1
-  and ia.not_before < $2
-  and (ia.not_after is null or ia.not_after > $2)
-`
+const resourceByPrivateIPQuery = "SELECT ia.private_ip, " +
```
sydneyteh96: For clarification, all sql queries are declared as string because constant cannot be passed as sql query.

pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
...
-from aws_public_ip_assignment ia
-         left join aws_resource res on ia.aws_resource_id = res.id
-         left join aws_region ar on res.aws_account_id = ar.id
-         left join aws_resource_type rt on res.aws_resource_type_id = rt.id
-         left join aws_account aa on res.aws_account_id = aa.id
-where ia.aws_hostname = $1
-  and ia.not_before < $2
-  and (ia.not_after is null or ia.not_after > $2)
-`
+const resourceByHostnameQuery = "SELECT ia.aws_hostname, " +
```
sydneyteh96: Same as the comment above.

pkg/storage/db.go
```diff
@@ -81,57 +81,56 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
 
 // Query to find resource by private IP using v2 schema
 // nolint
-const resourceByPrivateIPQuery = `select ia.private_ip,
```
sydneyteh96: Read clarification in comment below.

pkg/storage/db.go
sydneyteh96: Yes. I have fixed the queries.

sydneyteh96: Thanks for the catch.

pkg/storage/db.go
sydneyteh96: This is because it queries the old database, which is version 1. I think we need to update this function in a new PR. Thank you for asking.

ensary: ReadsFromNewSchemaVersion from https://github.com/asecurityteam/asset-inventory-api/pull/42/files is addressing this issue.

sydneyteh96: I will fix this PR to use ReadNewSchemaVersion.

pkg/storage/db.go
gcase555: question: weren't these already string literals?(the backtick ` indicates string literal) 

Why do they only need to be converted now if they were working as is? 

My source for thinking this doesn't need to be changed: https://stackoverflow.com/a/39025049

gcase555: personally I prefer backticks over needing to add `" +` to most lines of the string

sydneyteh96: I figured out that I had misunderstood the error which I encountered. In test, the ExpectQuery(query string) throws an error because of case sensitivity, so I have modified argument(change "SELECT" to "select"). Thank you for your explanation.

PR #41, opened 2020-02-10
Events backfill to shema V2 implementation

ensary:
* Back-fill via local calls. I was able to completely "backfill" data in my local Postgres in docker in less than one day, that looks good enough considering we are supposed to do it once
* Fixes to queries. There were some issues around using anonymous  SQL functions with numbered parameters, so I got rid of them and moved the logic of "on failed update insert" to go layer.
* Unique restrictions and indices where needed after query analysis
* Additional tests coming soon (I will update the PR)


gcase555:
nit: My first impression is this PR is rather large, clocking in at 500+ new lines.

We do have a lot of large PR's in our group, sometimes due to the config/code boilerplate needed to add new endpoints and handlers. Large PR's have felt odd to me since I got here, other projects I worked on fought hard to keep PR sizes small to ease reviews and lower risk.

I see PR needs all of the code here to satisfy the story, I just wonder if there was a way to break this up into a sequence of smaller changes.

I'm fine with it here but perhaps it's something we should discuss with the broader team.

pkg/handlers/v1/events_local_backfill_test.go
```diff
@@ -0,0 +1,109 @@
...
+		Runner: mockRunner,
+	}
+
+	from := time.Date(1970, 1, 0, 0, 0, 0, 0, time.UTC)
+	to := time.Date(2070, 1, 0, 0, 0, 0, 0, time.UTC)
+	err := handler.Handle(context.Background(), BackFillEventsInput{
+		From: from.Format(time.RFC3339Nano),
+		To:   to.Format(time.RFC3339Nano),
+	})
+	assert.Error(t, err)
```
gcase555: does this only have an error due to the mock returning one on https://github.com/asecurityteam/asset-inventory-api/pull/41/files#diff-9d2d11ac28b6c1b8f7104db5dce79de6R75 ?

pkg/storage/db.go
```diff
@@ -542,7 +543,7 @@ with sel as (
            aws_resource_type.id as aws_resource_type_id,
            val.meta
     from (
-             values (text $1, text $2, text $3, text $4, jsonb $5)
+             values ($1::text, $2::text, $3::text, $4::text, $5::jsonb)
```
gcase555: In case this threw anyone off, `::` is used for type casting in postgres https://stackoverflow.com/a/15537761 I had to double check that lol

pkg/storage/db.go
```diff
@@ -866,23 +882,107 @@ func (db *DB) releasePublicIP(ctx context.Context, tx *sql.Tx, arnID string, ip
...
-$$
-    begin
+	const releasePublicIPQueryUpdate = `
         update aws_public_ip_assignment
         set not_after=$1
         where public_ip = $2
           and aws_resource_id = (select id from aws_resource where arn_id = $3)
-          and aws_hostname = $4;
-        if not found then
+          and aws_hostname = $4`
```
dscgarcia: is it possible for there to be multiple records with the same public_ip, aws_resource_id, and aws_hostname (but different not_before/not_after values)? If so, might need a "where" clause to add "and not_after is null" so that the update doesn't overwrite all not_after values of other records

pkg/storage/db.go
```diff
@@ -790,25 +791,29 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
...
+set not_before = $1
+where private_ip = $2
+  and not_before = to_timestamp(0)
+  and not_after > $1
+  and aws_resource_id = (select id from aws_resource where arn_id = $3);`
+	const assignPrivateIPQueryInsert = `
+insert into aws_private_ip_assignment
+    (not_before, private_ip, aws_resource_id)
+values ($1, $2, (select id from aws_resource where arn_id = $3)) on conflict do nothing ;`
+	res, err := tx.ExecContext(ctx, assignPrivateIPQueryUpdate, when, ip, arnID)
```
gcase555: nit: a newline here(after the  sql) would help separate the sql strings from the go logic and improve readability

pkg/storage/db.go
```diff
@@ -24,7 +24,8 @@ const (
 	tableAWSHostnames          = "aws_hostnames"
```
gcase555: question: did we intend to break the previous format of putting all the sql queries as constant's above the functions like here https://github.com/asecurityteam/asset-inventory-api/blob/acd88d6f3a3698aa3d778be6caa04a7a9de571c9/pkg/storage/db.go#L85 ? right now we have both styles, sql outside of the functions and sql inside the functions.

pkg/storage/db.go
```diff
@@ -817,47 +822,58 @@ func (db *DB) releasePrivateIP(ctx context.Context, tx *sql.Tx, arnID string, ip
...
+update aws_private_ip_assignment
+set not_after=$1
+where private_ip = $2
+  and aws_resource_id = (select id from aws_resource where arn_id = $3);`
+
+	const releasePrivateIPQueryInsert = `
+insert into aws_private_ip_assignment
+    (not_before, not_after, private_ip, aws_resource_id)
+values (to_timestamp(0), $1, $2, (select id from aws_resource where arn_id = $3)) on conflict do nothing ;`
+	res, err := tx.ExecContext(ctx, releasePrivateIPQueryUpdate, when, ip, arnID)
```
gcase555: nit: ditto above https://github.com/asecurityteam/asset-inventory-api/pull/41/files#r377273349, newline here would improve readability

pkg/storage/db.go
gcase555: I feel like aws_resource_id's should be unique 🤔

pkg/handlers/v1/events_local_backfill_test.go
ensary: Yes, this is to make sure the handler throws the error is the storage back-end failed.

pkg/storage/db.go
ensary: Very good point, I will look into this.

pkg/storage/db.go
ensary: I would err on the side of constant inside the function if it is not used anywhere else. Does this sound acceptable?

gcase555: discussion we had:

Only SQL that is shared amongst the functions should be declared at the top of the file as a global type variable, SQL only used inside the functions themselves can stay inside the functions.

pkg/storage/db.go
ensary: Done.

api.yaml
```diff
@@ -500,7 +500,53 @@ paths:
                 #! end !#
                 "bodyPassthrough": true
             }
-
+  /ops/pgsql/v1/backfill:
+    get:
```
sydneyteh96: I suggest changing "get" to "post".

PR #43, opened 2020-02-19
Add reads (hostname) for new schema

dscgarcia:
Adding reads for hostname from new schema. Still need to test locally but wanted to get early feedback. 
- small edits to resourceByHostname query
- updated tests

Waiting on SID-219 and SID-220, I will pull those changes before merging this.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -753,7 +753,7 @@ func (db *DB) runQuery(ctx context.Context, version uint, query string, args ...
 		var isJoin bool      // no need for sql.NullBool as the DB column is guaranteed a value
 		var timestamp time.Time
 
-		if version < DualWritesSchemaVersion {
+		if version < ReadsFromNewSchemaVersion {
 			err = rows.Scan(&row.ARN, &ipAddress, &hostname, &isPublic, &isJoin, &timestamp, &row.AccountID, &row.Region, &row.ResourceType, &metaBytes)
 		} else {
 			err = rows.Scan(&ipAddress, &hostname, &row.ARN, &metaBytes, &row.Region, &row.ResourceType, &row.AccountID)
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"net"
 	"strings"
 	"sync"
 	"time"
@@ -50,8 +49,6 @@ const (
 	// ReadsFromNewSchemaVersion Lowest version of database schema that supports reads from M1 schema
 	ReadsFromNewSchemaVersion uint = 4
 )
-
-var privateIPs = []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"}
 
 // can't use Sprintf in a const, so...
 // %s should be `aws_hostnames_hostname` or `aws_ips_ip`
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -1003,3 +1024,17 @@ order by ae.ts asc
...
+		_, block, _ := net.ParseCIDR(privateIP)
+		privateIPBlocks = append(privateIPBlocks, block)
+	}
+	for _, block := range privateIPBlocks {
+		if block.Contains(ip) {
+			return false
+		}
+	}
+	return true
+}
```
dscgarcia: not crazy about this function having to parse the same array of private IP addresses every single time a read comes in for the new schma but wasn't sure how else to do this without some sort of init() function. Open to suggestions :)

pkg/storage/db.go
```diff
@@ -701,22 +705,32 @@ func (db *DB) insertNetworkChangeEvent(ctx context.Context, timestamp time.Time,
...
 // FetchByHostname gets the assets who have hostname at the specified time
 func (db *DB) FetchByHostname(ctx context.Context, when time.Time, hostname string) ([]domain.CloudAssetDetails, error) {
-	sqlstmt := fmt.Sprintf(latestStatusQuery, `aws_hostnames_hostname`)
-	return db.runQuery(ctx, sqlstmt, hostname, when)
+	ver, err := db.GetSchemaVersion(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var assets []domain.CloudAssetDetails
+	if ver < DualWritesSchemaVersion {
```
ensary: We have a separate version for reads from new schema.

pkg/storage/db.go
```diff
@@ -739,7 +753,14 @@ func (db *DB) runQuery(ctx context.Context, query string, args ...interface{}) (
 		var isJoin bool      // no need for sql.NullBool as the DB column is guaranteed a value
 		var timestamp time.Time
 
-		err = rows.Scan(&row.ARN, &ipAddress, &hostname, &isPublic, &isJoin, &timestamp, &row.AccountID, &row.Region, &row.ResourceType, &metaBytes)
+		if version < DualWritesSchemaVersion {
+			err = rows.Scan(&row.ARN, &ipAddress, &hostname, &isPublic, &isJoin, &timestamp, &row.AccountID, &row.Region, &row.ResourceType, &metaBytes)
+		} else {
+			err = rows.Scan(&ipAddress, &hostname, &row.ARN, &metaBytes, &row.Region, &row.ResourceType, &row.AccountID)
+			ip, _, _ := net.ParseCIDR(ipAddress)
+			isPublic = isPublicIP(ip)
```
ensary: Private IPs are in different table. The returned by this query is always public.

pkg/storage/db.go
dscgarcia: Good catch, will update that

pkg/storage/db.go
dscgarcia: True. I was preparing for when we implement resourceByPrivateIPQuery, I assume what it will return will be mapped to the same variable name as the public ip query ("ipAddress"). 
But I can change it to be static (isPublic = true) on this PR if that keeps things cleaner.

pkg/storage/db.go
```diff
@@ -50,6 +51,8 @@ const (
 	ReadsFromNewSchemaVersion uint = 4
 )
 
+var privateIPs = []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"}
```
ensary:
```
var A  = net.IPNet{
	IP: net.IPv4(192, 168, 1, 0)  ,
	Mask: net.IPv4Mask(255, 255, 255,0 ),
}
```
as an alternative?

dscgarcia: perfect 👍

PR #44, opened 2020-02-19
Update fetch assets by private and public IP to match M1 schema

sydneyteh96:
"Find by Public IP" and "Find by Private IP" are combined into one pull request because the code change affects both. Summary of pull request:
- Add hostname as one of the return values in _resourceByPublicIPQuery_ query
- FetchByIP() function modified to work with old and new schema
- Add runFetchByIPQuery() to run queries, which are _resourceByPublicIPQuery_ and _resourceByPrivateIPQuery_ on new schema when called
- Add test cases

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -87,39 +87,40 @@ const latestStatusQuery = "WITH latest_candidates AS ( " +
 
 // Query to find resource by private IP using v2 schema
 // nolint
-const resourceByPrivateIPQuery = "SELECT ia.private_ip, " +
-	"			res.arn_id, " +
-	"			res.meta, " +
-	"			ar.region, " +
-	"			rt.resource_type, " +
-	"			aa.account " +
-	"	FROM aws_private_ip_assignment ia " +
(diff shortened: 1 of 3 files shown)


pkg/storage/db.go
```diff
@@ -712,8 +714,128 @@ func (db *DB) FetchByHostname(ctx context.Context, when time.Time, hostname stri
...
+	ver, err := db.GetSchemaVersion(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var asset []domain.CloudAssetDetails
+	if ver < ReadsFromNewSchemaVersion {
+		sqlstmt := fmt.Sprintf(latestStatusQuery, `aws_ips_ip`)
+		asset, err = db.runQuery(ctx, sqlstmt, ipAddress, when)
+	} else {
+		ipaddr := net.ParseIP(ipAddress)
```
ensary: ParseIP returns nil if the string passed to it is not a valid IP. Can you please add a check for this?

sydneyteh96: Yes, I have added a check. Thank you for catching.

PR #46, opened 2020-02-26
Workaround for using closed connections

ensary:
The schema management code uses postgres driver directly w/o golang sql wrapper, so it does not have reconnect logic that is present in go sql by default:
https://github.com/golang/go/blob/master/src/database/sql/sql.go#L784
Adding PingContext should re-initialize connection if it is in bad state before any schema related queries are run.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -242,7 +242,10 @@ func (db *DB) GetSchemaVersion(ctx context.Context) (uint, error) {
 }
 
 func (db *DB) migrateSchema(ctx context.Context, d migrationDirection) (uint, error) {
-	var err error
+	err := db.sqldb.PingContext(ctx)
+	if err != nil {
+		return 0, err
+	}
 	switch d {
 	case up:
 		err = db.migrator.Steps(1)
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -196,6 +196,12 @@ var privateIPNetworks = []net.IPNet{
 
 // ForceSchemaToVersion sets the database schema to specified version without running any migrations and clears dirty flag
 func (db *DB) ForceSchemaToVersion(ctx context.Context, version uint) error {
+	// migrator is using raw Postgres connection which lacks re-try logic
+	// ensure we have an established connection before running any migrate commands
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -211,11 +217,19 @@ func (db *DB) MigrateSchemaDown(ctx context.Context) (uint, error) {
...
+	}
 	return db.migrator.Migrate(version)
 }
 
 // GetSchemaVersion retrieves the current version of database schema
 func (db *DB) GetSchemaVersion(ctx context.Context) (uint, error) {
+	err := db.sqldb.PingContext(ctx)
+	if err != nil {
+		return 0, err
+	}
```
dscgarcia: do we need this change in migrateSchema() func as well (line 245, called by MigrateSchemaUp/MigrateSchemaDown)?

pkg/storage/db_test.go
```diff
@@ -1429,8 +1453,14 @@ func TestForceSchemaToVersionSuccess(t *testing.T) {
 	version := uint(rand.Uint64())
 	migrator := NewMockStorageMigrator(ctrl)
 	migrator.EXPECT().Force(int(version)).Return(nil)
-	db := &DB{migrator: migrator}
-	err := db.ForceSchemaToVersion(context.Background(), version)
+	mockdb, mock, err := sqlmock.New()
```
gcase555: nit: this feels quite repetitive(not DRY) although I noticed the rest o0f the tests in this file have significant duplication as well. A future improvement would be creating some kind of shared testing setup and fixtures.

ensary: We historically have lower expectations for code quality in tests (AFAIU the linting does not cover them much either), so it is a good thing to think about improvements. TBH I was thinking the same thing while copy/pasting this code :)

gcase555: Probably needs to be broken up into multiple files as well, any code file above 1000+ lines is an immediate code smell

ensary: We are about to drop half of all the code and tests in db as we will remove old schema support. But even after that, I agree - migrations could probably be extracted for example.

PR #47, opened 2020-03-04
Fix reconnect issues, fix schema version on replica, simplify and clean up DB layer

ensary:
* Use shared configuration for primary and replica, remove replica-specific config
* Simplify configuration of DB connection by using URL instead of separate parameters that were not needed
* Simplify connection logic by removing handling of TLS to configuration (url vs auto-detection) and some additional re-connects that were leftover of manual database creation
* Allow fall-back to primary if replica is not configured (for docker-compose, integration tests etc.)
* Move schema management out of DB struct into separate file (schema) and struct (SchemaManager)
* Use separate connections and different reconnect logic for schema management commands exposed via API
* Add optimized getSchemaVersion to DB that bypasses the golang/migrate to avoid connection re-use and related performance issues
* Refactor integration tests to work with the changes above


pkg/storage/config.go
```diff
@@ -2,25 +2,26 @@ package storage
...
 
 // PostgresConfig contains the Postgres database configuration arguments
 type PostgresConfig struct {
-	Hostname         string
-	Port             uint16
-	Username         string
-	Password         string
-	DatabaseName     string
+	URL              string
+	ReplicaURL       string
```
dscgarcia: do we need to add the replica url as an env var?

ensary: Yes, but it is optional now. I.e. the code falls back to using primary for everything is the replica url is not set or is not valid/reachable.

PR #49, opened 2020-05-07
Adding champions to account owners

gcase555:
We are modifying the database to satisfy the following constraints:
```
Every account has only one owner

Every account has one or more champions

Any of champions can be null (but should not)
```

**TODO:**
- [x] add table for keeping track of account owners.
- [x] add unique constraints to person table
- [x] test create queries work when run during a migration
- [x] task: add indexes, at least for the fields that will be queried frequently
- [x] test manually

### DIFF
--- a/db-migrations/000005_adding_owners.up.sql
+++ b/db-migrations/000005_adding_owners.up.sql
@@ -10,15 +10,19 @@ create table person(
 );
 
 create table account_champion(
+    person_id serial,
+    aws_account_id integer,
     foreign key(person_id) references person (id),
     foreign key(aws_account_id) references aws_account (id),
-    constraint account_champ_unique unique(person_id, aws_account_id)
+    unique(person_id, aws_account_id)
 );
 
 create table account_owner(
+    person_id serial,
+    aws_account_id integer,
     foreign key(person_id) references person (id),
     foreign key(aws_account_id) references aws_account (id),
-    constraint account_unique unique (aws_account_id)
+    unique (aws_account_id)
 );
 
 COMMIT;
--- a/db-migrations/000005_adding_owners.up.sql
+++ b/db-migrations/000005_adding_owners.up.sql
@@ -5,9 +5,14 @@ create table champions(
     champion varchar not null
 )
 
+create table owners(
+    id serial primary key,
+    owner varchar not null,
+    foreign key(aws_account_id) references aws_account (id)
+)
+
 alter table aws_account
 -- accounts may or may not have champions associated with them, so we may not need to enforce a foreign key restraint
 add column champion_id serial null unique
-add column owner varchar unique
 
 COMMIT;
--- a/db-migrations/000005_adding_owners.up.sql
+++ b/db-migrations/000005_adding_owners.up.sql
@@ -8,5 +8,6 @@ create table champions(
 alter table aws_account
 -- accounts may or may not have champions associated with them, so we may not need to enforce a foreign key restraint
 add column champion_id serial null unique
+add column owner varchar unique
 
 COMMIT;
--- a/db-migrations/000005_adding_owners.up.sql
+++ b/db-migrations/000005_adding_owners.up.sql
@@ -1,0 +1,3 @@
+-- schema changes for account owners and champions
+BEGIN;
+COMMIT;
--- a/db-migrations/000005_adding_owners.up.sql
+++ b/db-migrations/000005_adding_owners.up.sql
@@ -2,7 +2,7 @@
 BEGIN;
 create table champions(
     id serial primary key,
-    champion varchar not null,
+    champion varchar not null
 )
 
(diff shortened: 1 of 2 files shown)


gcase555:
I am realizing I may need to add a many to many table to satisfy:
- `Every account has one or more champions`

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,13 @@
+-- schema changes for account owners and champions
```
gcase555, line 1: TODO: add a owners table

gcase555: FYI this is being refactored now

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,22 @@
...
+create table champions(
+    foreign key(people_id) references person (id),
+    foreign key(aws_account_id) references aws_account (id)
+)
+
+create table owner(
+    foreign key(people_id) references person (id),
+    foreign key(aws_account_id) references aws_account (id)
+)
+
```
gcase555, line 21: task: add indexes, at least for the fields that will be queried frequently e.g. name, email

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,22 @@
+-- schema changes for account owners and champions
```
gcase555, line 1: @ensary this is a many-many design similar to what we discussed, one aspect of this though is that owner is technically `one to many`, since an account can have only one owner. This is not currently modeled with these many-many tables and I'll need to enforce. 

I am considering adding a FK owner column to the aws_accounts table but one problem with that will be populating that column and is going to be more difficult in some ways.

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,22 @@
...
+BEGIN;
+
+create table person(
+    id serial primary key,
+    login varchar not null,
+    email varchar not null,
+    name varchar not null,
+    valid boolean
```
gcase555, line 9: debating if this should be null or even including, the idea of a person being `valid` is a bit odd

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,22 @@
...
+)
+
+create table champions(
+    foreign key(people_id) references person (id),
+    foreign key(aws_account_id) references aws_account (id)
+)
+
+create table owner(
```
gcase555, line 17: `owners` would be a better name if its a many-many table

db-migrations/000005_adding_owners.up.sql
gcase555: `Any of owner/champions can be null (but should not)`
Makes me not want to use a foreign key though

db-migrations/000005_adding_owners.up.sql
ensary: We can turn this into owner_id in the accounts table as account can have only one owner.

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,22 @@
...
+
+create table person(
+    id serial primary key,
+    login varchar unique not null,
+    email varchar unique not null,
+    name varchar not null,
+    valid boolean not null
+)
+
+create table champions(
```
gcase555, line 12: TODO: figure out how to guarantee each champion-account pair is unique

gcase555: can probably do this with a constraint

db-migrations/000005_adding_owners.up.sql
```diff
@@ -0,0 +1,23 @@
...
+    foreign key(people_id) references person (id),
+    foreign key(aws_account_id) references aws_account (id),
+    constraint champ_unique unique(people_id, aws_account_id)
```
gcase555, line 15: trying to use a constraint here as per https://stackoverflow.com/questions/10468657/postgres-unique-multi-column-index-for-join-table to guarantee we get unique pairs of people and accounts

gcase555:
I wasn't sure about automated tests at this stage, it'd seems better to test this as part of an integration test later if we wanted to automate checking creating persons and associating them as champions.

I tested this manually by creating a container with this code, building it with the bitbucket asset-inventory-api, running migrations and manually inserting records into the schema using test data in `psql`.

dscgarcia:
I know we discussed having the owners table separate from aws_account as a temporary solution to backfill owner/person ids, and that in the future we would create another migration script to move the owner data as a foreign key column into the aws_account table. Are we tracking that in a future ticket or this one?

gcase555:
@dscgarcia ticket has been created and is in the M2 epic. I'll link it in our slack room.

PR #51, opened 2020-05-08
Stopgap script to unblock work depending on resources export for point in time

ensary:
**Caveat: this is not the code I am most proud of. It solves the problem of bulk export with new schema that is a blocker right now, and which we can not implement properly due to conflicting priorities. The expectation is for "proper" bulk export API with new schema to be implemented later.**

The script talks directly to the Postgres DB (ideally - replica). All JSON serialization is done on Postgres side. Some local wrangling with `jq` to combine results.

Output is designed to be 100% compatible with output of API endpoint described in api.yaml, so any existing code to parse the data can be re-used.

I was able to run this with staging DB instance that has large amount of data, and it produces valid JSON that has the number of entries matching the count of active resources for point in time in the database.

### DIFF
--- a/bulk-export/README.md
+++ b/bulk-export/README.md
@@ -1,0 +1,31 @@
+# Bulk export of resource assignments for point in time
+## Summary
+The [shell script](export.sh) and [SQL file](export.sql) allow to export all the resources 
+and assignments for the specified timestamp utilizing new database schema.
+
+The export runs manually via command line using connection to the Postgres database. 
+This is a __STOPGAP__ solution until the bulk export endpoint `/v1/cloud/asset` is updated to use the new SQL schema.
+
+The resulting JSON dump file has the format identical to the one defined in `#/components/schemas/BulkCloudAssets` in the 
+[API description](../api.yaml)
+
+## Using it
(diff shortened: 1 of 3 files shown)


gcase555:
design questions:
- how do we plan on running this on a regular basis? cron man/job or something like that?
- how will the status of execution of the 'bulk export' be bubbled up to concerned consumers?

ensary:
> design questions:
> 
>     * how do we plan on running this on a regular basis? cron man/job or something like that?
> 
>     * how will the status of the execution of this 'bulk export' script be bubbled up to concerned consumers?
Current plan is to run manually on demand and give the resulting export to consumers. 
We can discuss running this via cron  after we make sure it works well enough. RE: notification - I guess putting the dump on S3 and sending a message re: dump being ready and path to SNS topic could be an option.

gcase555:
Ran `shellcheck` against the shell script to see if I got any hits but surprisingly no, well done.

gcase555:
We might need to supplement the README with a runbook to capture things that are more team/Atlassian specific but the code here LGTM as a solution until we get to M4.

PR #52, opened 2020-05-18
adding cascade statement

gcase555:
Step down needs a `cascade` in order to delete the dependent objects


gcase555:
Tested this out locally and was able to drop the listed tables successfully.

PR #53, opened 2020-05-18
Add new endpoint that updates or inserts account owner, account ID and champions

sydneyteh96:


pkg/storage/db.go
```diff
@@ -1058,3 +1058,84 @@ order by ae.ts asc
...
+}
+
+func (db *DB) storeAccountOwner(ctx context.Context, accountOwner domain.AccountOwner, tx *sql.Tx) error {
+
+	sqlStatement := fmt.Sprintf(`REPLACE INTO aws_account (account) VALUES($1)`) // nolint
+	if _, error := tx.ExecContext(ctx, sqlStatement, accountOwner.AccountID); error != nil {
+		return error
+	}
+
+	sqlStatement = fmt.Sprintf(`IF EXISTS(SELECT * from person WHERE login=$1)
```
gcase555, line 1092: task: we should probably move the multi-line queries into constants like up above https://github.com/asecurityteam/asset-inventory-api/blob/85b1bc58ae14d761cb9f95c29ab968be27db6925/pkg/storage/db.go#L147

pkg/storage/db.go
```diff
@@ -1058,3 +1058,84 @@ order by ae.ts asc
...
+									 UPDATE person SET email=$2, name=$3, valid=$4
```
ensary, line 1093: This part seems to be missing "where" clause, and will update values for all entries in the `person` table

bulk-export/README.md
```diff
@@ -0,0 +1,31 @@
+# Bulk export of resource assignments for point in time
```
gcase555, line 1: question: why is the bulk-export code in this pull-request?

gcase555: FYI: it appears the diff is showing other commits, we should probably try to rebase instead of merge commit to hide these.

sydneyteh96:
New pull request that does not include old commits that are merged into masters  https://github.com/asecurityteam/asset-inventory-api/pull/55

PR #55, opened 2020-06-01
Add new endpoints that updates or inserts aws accounts, owners and champions

sydneyteh96:


pkg/handlers/v1/insert_account_owner_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func newInsertAccountOwnerHandler(storer domain.AccountOwnerStorer) *AccountOwnerInsertHandler {
+	return &AccountOwnerInsertHandler{
+		LogFn:              testLogFn,
+		StatFn:             testStatFn,
+		AccountOwnerStorer: storer,
+	}
+}
+
+func validInput() AccountOwner {
```
gcase555, line 21: info: `validInput()` is slightly confusing as name, it made me think of input validation.

Could we name this something slightly more descriptive to testing? `testAccountOwner`, `testInput` or `testData`?

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/gobuffalo/genny v0.1.1 // indirect
 	github.com/gobuffalo/gogen v0.1.1 // indirect
 	github.com/golang-migrate/migrate/v4 v4.8.0
-	github.com/golang/mock v1.4.0
+	github.com/golang/mock v1.4.3
```
sydneyteh96, line 19: Installed latest mockgen package to generate code.

pkg/handlers/v1/insert_account_owner_test.go
sydneyteh96: Yes, sure.

api.yaml
```diff
@@ -582,6 +582,40 @@ paths:
...
+                $ref: "#/components/schemas/Error"
+      x-transportd:
+        backend: app
+        enabled:
+          - "accesslog"
+          - "requestvalidation"
+          - "responsevalidation"
+          - "lambda"
+        lambda:
+          arn: "updateAccountOwner"
```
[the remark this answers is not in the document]
gcase555: I believe you are correct @alecbert-atl the other api endpoints match the lamdba arn to the handler that is declared in main.go https://github.com/asecurityteam/asset-inventory-api/pull/55/files#diff-7ddfb3e035b42cd70649cc33393fe32cR145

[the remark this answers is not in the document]
sydneyteh96: Yes, you are correct. I forgot to change lambda arn after modifying handler name. Thank you for catching the error.

pkg/storage/db.go
```diff
@@ -1060,3 +1083,82 @@ order by ae.ts asc
...
+		return err
+	}
+
+	return nil
+}
+
+func (db *DB) storeAccountOwner(ctx context.Context, accountOwner domain.AccountOwner, tx *sql.Tx) error {
+
+	sqlStatement := fmt.Sprintf(`REPLACE INTO aws_account (account) VALUES($1)`) // nolint
+	if _, error := tx.ExecContext(ctx, sqlStatement, accountOwner.AccountID); error != nil {
```
gcase555, line 1113: we could just use one `error` instead of declaring different ones here like `error1`, `error2`

sydneyteh96: Done

pkg/storage/db.go
```diff
@@ -1060,3 +1083,83 @@ order by ae.ts asc
...
+	if err = tx.Commit(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (db *DB) storeAccountOwner(ctx context.Context, accountOwner domain.AccountOwner, tx *sql.Tx) error {
+
+	sqlStatement := fmt.Sprintf(`REPLACE INTO aws_account (account) VALUES($1)`)
```
dscgarcia, line 1112: I've never seen REPLACE INTO before, is this like an upsert? It creates if it doesn't exist and updates if it does?

sydneyteh96: It works like upsert, except it will delete the values of old records and insert new values, if primary key or unique key in old records are the same as new records.

sydneyteh96: Actually PostgresSQL does not have REPLACE INTO, and MySQL has it. Upsert is a way to do the job. I will make changes to the query. Thank you for your comment.

pkg/storage/db.go
```diff
@@ -1060,3 +1083,83 @@ order by ae.ts asc
...
+		return err
+	}
+	var id int
+	row.Next()
+	if err := row.Scan(&id); err != nil {
+		return err
+	}
+
+	sqlStatement = fmt.Sprintf(insertAccountOwnerQuery, id)
+	if _, err := tx.ExecContext(ctx, sqlStatement, accountOwner.AccountID); err != nil {
```
alecbert-atl, line 1134: Question: Why does the insertAccountOwnerQuery mix ways to add arguments? I would have thought you could insert id and accountOwner.AccountID into the query with ExecContext, so I wanted to check if there's a limitation I'm missing since I haven't used it before.

If this is simply a style thing please disregard.

pkg/storage/db.go
```diff
@@ -1109,34 +1093,41 @@ func (db *DB) StoreAccountOwner(ctx context.Context, accountOwner domain.Account
...
+	sqlStatement = fmt.Sprintf(`SELECT id FROM aws_account WHERE account=$1`)
+	row = tx.QueryRowContext(ctx, sqlStatement, accountOwner.AccountID)
+	var accountID int
+	if err := row.Scan(&accountID); err != nil {
 		return err
 	}
 
-	sqlStatement = fmt.Sprintf(insertAccountOwnerQuery, id)
-	if _, err := tx.ExecContext(ctx, sqlStatement, accountOwner.AccountID); err != nil {
+	sqlStatement = fmt.Sprintf(`INSERT INTO account_owner VALUES %d, %d`, personID, accountID)
```
dscgarcia, line 1124: Should this query also handle updates to account_owner? for example, if the account_id already exists in the table but there is a new person_id assigned to it

sydneyteh96: Yes. When there is another pair of accoun_id and person_id from input, the pair will be inserted into the table if the account_id already exists. The unique key in the table is (account_id, person_id) so the table will be inserted or updated, except there is input with existing account_id and person_id pair.

sydneyteh96: No. This is because values in aws_account_id column should be unique according to account_owner table. I am fixing the query. Thank you for your question.

sydneyteh96: Done

gcase555:
@sydneyteh96 I tested that the `test.json` file worked with the following command:
`curl -H "Content-Type: application/json"  http://localhost:8080/v1/account/owner -d @test.json` with my latest commit.

I also fixed some merge conflicts around moving a query back up to the top of the file, you will likely want to pull before trying to do any more work.

pkg/storage/db.go
```diff
@@ -1060,3 +1067,89 @@ order by ae.ts asc
...
+			if _, err := tx.ExecContext(ctx, insertPersonQuery, person.Login, person.Email, person.Name, person.Valid); err != nil {
+				return err
+			}
+
+			row := tx.QueryRowContext(ctx, `SELECT id FROM person WHERE login=$1`, person.Login)
+			var champID int
+			if err := row.Scan(&champID); err != nil {
+				return err
+			}
+			sqlStatement = fmt.Sprintf(`INSERT INTO account_champion(person_id, aws_account_id) VALUES (%d, %d) ON CONFLICT DO NOTHING`, champID, accountID)
```
gcase555, line 1148: task: can we change this query to just use the `$1` style instead of using `fmt.Sprintf` to substitute the variable

pkg/storage/db.go
```diff
@@ -1060,3 +1067,89 @@ order by ae.ts asc
...
+	}
+
+	sqlStatement = `INSERT INTO account_owner
+		VALUES ($1, $2)
+   		ON CONFLICT (aws_account_id) DO UPDATE SET person_id = $1, aws_account_id = $2`
+	if _, err := tx.ExecContext(ctx, sqlStatement, personID, accountID); err != nil {
+		return err
+	}
+
+	sqlStatement = fmt.Sprintf(`DELETE FROM account_champion WHERE aws_account_id=%d`, accountID)
```
gcase555, line 1132: task: can we change this query to just use the `$1` style instead of using `fmt.Sprintf` to substitute the variable

test.json
```diff
@@ -0,0 +1,23 @@
+{
```
gcase555, line 1: task: we used this for testing purposes but you can probably delete it for now. If we're going to store test data in the repo, it's better we do it as a test fixture and have a test use it.

sydneyteh96: fixed

pkg/storage/db.go
sydneyteh96: Fixed

pkg/storage/db.go
```diff
@@ -1060,3 +1067,96 @@ order by ae.ts asc
...
+		return err
+	}
+
+	return nil
+}
+
+func (db *DB) storeAccountOwner(ctx context.Context, accountOwner domain.AccountOwner, tx *sql.Tx) error {
+
+	sqlStatement := `
+				INSERT INTO aws_account (account)
```
gcase555, line 1097: nit: formatting looks slightly off here? looks like slightly too many tabs?

sydneyteh96: done

pkg/storage/db.go
sydneyteh96: Fixed

PR #58, opened 2020-06-17
Add endpoint to fetch cloud assets by arn ID

sydneyteh96:
New endpoint is added to fetch cloud asset, its account owner and champions by ARN ID.  
- There are two database queries. The '"resourceByARNID" query gets cloud asset details with metadata by ARN ID and timestamp. The other query, "ownerByAccountID" finds account owner and champions by account ID, which is an auto-increment primary key associated with an actual account ID.
- A different function is created to run "ownerByAccountID" query and return a struct that has account ID, owner and champions.   
- The existing integration tests are modified due to changes in domain.cloud.
- A ticket is created for updating "find by ip" and "find by hostname" to return account owner and champions.


api.yaml
```diff
@@ -754,7 +814,7 @@ components:
     AccountOwner:
       type: object
       properties:
-        acountId:
+        accountId:
           type: string
```
sydneyteh96, line 818: This is a typo fix. The existing endpoints works as expected with the typo.

api.yaml
```diff
@@ -322,7 +322,65 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
-
+  /v1/cloud/arnid/{arnid}:
```
rsoni-atlassian, line 325: do we need resource specific to arnid? can we have path /v1/cloud/{arnid} instead?

api.yaml
```diff
@@ -322,7 +322,65 @@ paths:
...
+    get:
+      summary: "Retrieve a cloud asset at a point in time by ARN ID"
```
rsoni-atlassian, line 327: nit: is time, current time?

api.yaml
```diff
@@ -322,7 +322,65 @@ paths:
...
+      parameters:
+        - name: "arnid"
```
rsoni-atlassian, line 329: nit: Please confirm arnid pattern doesn't break url or require character escaping

api.yaml
```diff
@@ -322,7 +322,65 @@ paths:
...
+          in: "path"
+          description: "The arn id of the asset"
+          required: true
+          schema:
+            type: "string"
+        - name: "time"
```
rsoni-atlassian, line 335: nit: may be call timeStamp?

api.yaml
```diff
@@ -322,7 +322,65 @@ paths:
...
+          in: "query"
+          description: "The point in time details for a given asset"
+          required: true
+          schema:
+            type: "string"
+            format: "date-time"
```
rsoni-atlassian, line 341: It seems like implementation is using RFC3339Nano which is very specific time formatting. Do we expect consumers of API to understand this formatting? If so, i suggest we communicate with them in more details.

api.yaml
sydneyteh96: From my understanding, yes because there are other endpoints that find resources by IP address and hostname. This path can be used to find resources specific to an ARN ID.

pkg/handlers/v1/cloud_fetch.go
```diff
@@ -31,14 +31,15 @@ type PagedCloudAssets struct {
...
-	Tags               map[string]string `json:"tags"`
+	PrivateIPAddresses []string            `json:"privateIpAddresses"`
+	PublicIPAddresses  []string            `json:"publicIpAddresses"`
+	Hostnames          []string            `json:"hostnames"`
+	ResourceType       string              `json:"resourceType"`
+	AccountID          string              `json:"accountId"`
+	Region             string              `json:"region"`
+	ARN                string              `json:"arn"`
+	Tags               map[string]string   `json:"tags"`
+	AccountOwner       domain.AccountOwner `json:"accountOwner"`
```
ensary, line 42: This is missing list of account champions.

ensary, line 42: The api descriptor too (missing champions).

gcase555: +1 to this please fix @sydneyteh96

api.yaml
gcase555: We generally tend to use the api.yaml when looking up how to make requests to endpoints, a comment here indicating this time is in `RFC3339Nano` form would be helpful as a start. Eventually this information should exposed in the api docs for consumers to be aware of.

pkg/handlers/v1/cloud_fetch.go
sydneyteh96: Champions is a field in AccountOwner struct.

gcase555: got it!

gcase555:
Mostly looks good to me, you do have failing builds to deal with though.

api.yaml
sydneyteh96: The ARN ID contains ASCII safe characters. For example, "i-123423453b56jdk806". So, it does not require character escaping.

api.yaml
sydneyteh96: I have changed the name.

sydneyteh96:
> Mostly looks good to me, you do have failing builds to deal with though.

@gcase555 The changes broke existing tests in test/storage_test.go. I have modified the tests so it is fixed.

api.yaml
gcase555: @sydneyteh96 could be useful to add a comment that the date-time is `RFC3339Nano` format but I'll leave it up to you, it could technically be done in the API docs.

sydneyteh96: Yes, I can add a comment.

pkg/storage/db_test.go
```diff
@@ -732,6 +732,303 @@ func TestGetHostnamesAtTimeMultiRowsSchema2(t *testing.T) {
...
+
+	thedb := DB{
+		sqldb: mockdb,
+	}
+
+	withSchemaVersion(4, mock)
+	at, _ := time.Parse(time.RFC3339, "2019-04-09T08:55:35+00:00")
+	const arnID = "arnid"
+
+	rows := sqlmock.NewRows([]string{"aws_private_ip_assignment_private_ip",
```
gcase555, line 750: info: I could be beneficial to try to reduce the amount of duplication here. These test rows are created 3x times and then populated with different testing data depending on the scenario, sometimes shared data and sometimes different. You might consider using a technique like table driven tests https://github.com/golang/go/wiki/TableDrivenTests to reduce duplication in the tests.

gcase555: I am open to this being improved as a follow up ticket/PR should you feel the scope/time on this ticket is already larger than you intended, also if for some reason table based tests would not work here then the duplication might be ok.

sydneyteh96: As all inputs for the test cases are the same, I think table driven tests will not work in this case. Instead, I can remove duplicate by creating a global variable, if you prefer.

gcase555: This will likely we fixed in a separate change as table based tests had issues in this spot.

sydneyteh96:
> Please revert the `time` argument to it's original name. This is a change that breaks the API.

@ensary done

api.yaml
alecbert-atl: Would this not change the api contract? If the current `/v1/account/owner` endpoint expects to have an object submitted with an "acountId", will changing this force users of that endpoint to change the object being POSTed?

sydneyteh96: The json declaration of "accountId" is here: https://github.com/asecurityteam/asset-inventory-api/blob/master/pkg/handlers/v1/insert_account_owner.go#L12. The endpoint should work with "accountId".

sydneyteh96: "acountId" does not match "accountId" in https://github.com/asecurityteam/asset-inventory-api/blob/master/pkg/handlers/v1/insert_account_owner.go#L12.

alecbert-atl: If AccountOwner is not in use yet this also makes sense to me.

sydneyteh96: The /v1/account/owner endpoint works with and without the typo fix, but I decided to fix it to match the struct in insert_account_owner.go.

alecbert-atl: Ooops, thank you! And I'm sorry about the follow up comment, I thought I had refreshed the page to see if you had replied but I guess I hadn't!

sydneyteh96: Thank you for asking so I can clarify

PR #62, opened 2020-07-08
Adding account owner and champion in fetchByHostname and fetchByIp re…

rsoni-atlassian:
…sponse

### DIFF
--- a/pkg/storage/db_test.go
+++ b/pkg/storage/db_test.go
@@ -319,13 +319,35 @@ func TestGetPrivateIPsAtTimeM1Schema(t *testing.T) {
 		"aws_resource_meta",
 		"aws_region_region",
 		"aws_resource_type_resource_type",
-		"aws_account_account"}).AddRow("10.2.2.6",
+		"aws_account_account",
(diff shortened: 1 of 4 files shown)


pkg/storage/db_test.go
```diff
@@ -319,13 +319,35 @@ func TestGetPrivateIPsAtTimeM1Schema(t *testing.T) {
...
+		"aws_account_account",
+		"aws_account_id"}).AddRow("10.2.2.6",
 		"rid",
 		[]byte("{\"hi\":\"there1\"}"),
 		"region",
 		"type",
-		"aid")
+		"aid",
+		"1")
+	rows2 := sqlmock.NewRows([]string{"aws_account_account",
```
rsoni-atlassian, line 330: self review: i noticed there is lot of improvement room wrt to how we are generating test data and using mock. if there is enough agreement, i will open task to refactor test code and reduce duplicate.

sydneyteh96: There is a hardening epic ticket about reducing duplicate. You can add description and edit the ticket or add a new ticket.

PR #64, opened 2020-07-21
Fix invalid result for empty lookup

ensary:
The code before this change used to return JSON with all fields but ARN ID empty for cases when non-existent ARN was passed. Fixed to return 404 instead.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -894,8 +894,9 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 	var asset domain.CloudAssetDetails
 	var accountID int
 	hasTag := false
+	empty := true
 	for rows.Next() {
-
+		empty = false // there is no way to check for size of SQL resul tset in golang
 		var privateIPAddress sql.NullString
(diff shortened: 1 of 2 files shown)


pkg/storage/db.go
```diff
@@ -894,8 +894,9 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 	var asset domain.CloudAssetDetails
 	var accountID int
 	hasTag := false
+	empty := true
 	for rows.Next() {
-
+		empty = false // there is no way to check for size of SQL resul tset in golang
```
gcase555, line 899: task: typo in this comment, should be `result set`

pkg/storage/db.go
```diff
@@ -894,8 +894,9 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 	var asset domain.CloudAssetDetails
 	var accountID int
 	hasTag := false
+	empty := true
 	for rows.Next() {
-
+		empty = false // there is no way to check for size of SQL result set in golang
```
gcase555, line 899: 👌 comment is helpful here

PR #65, opened 2020-07-22
Creating functions to leverage postgres performance

rsoni-atlassian:
### Overview

[SID-477](https://asecurityteam.atlassian.net/browse/SID-477) calls for investigation in ways we can improve performance of our queries.
As first step moving queries to functions, as we were running multiple SQL queries during single API calls for some APIs. This should give some performance improvement out of box.
Having queries part of function will allow us to profile function performance and optimize them without affecting service business layer.

#### On Local

- Published OSS image to local registry
- Consumes local OSS image in BB repository
- Invoked performance tests

#### On Dev

- Marking this PR as WIP to perform Dev perf testing


rsoni-atlassian:
Working on updating unit tests, this will also fix build failures

db-migrations/000007_adding_functions.down.sql
```diff
@@ -0,0 +1,10 @@
+-- schema changes to support dual writes
+BEGIN;
+
+DROP FUNCTION get_resource_by_arn_id(character varying, timestamp without time zone);
```
ensary, line 4: NIT: adding "IF EXISTS" to drop  raises our chances of recovering from botched schema (i.e. somehow some of the functions are defined but others are not, and we want to roll back)

db-migrations/000007_adding_functions.up.sql
```diff
@@ -0,0 +1,275 @@
+-- schema changes to support dual writes
+BEGIN;
+
+CREATE OR REPLACE FUNCTION get_resource_by_arn_id(aid VARCHAR, ts TIMESTAMP)
```
ensary, line 4: 👍 re: OR REPLACE

pkg/storage/db.go
```diff
@@ -65,115 +65,19 @@ const latestStatusQuery = `WITH latest_candidates AS (
...
-from aws_private_ip_assignment ia
-         left join aws_resource res on ia.aws_resource_id = res.id
-         left join aws_region ar on res.aws_region_id = ar.id
-         left join aws_resource_type rt on res.aws_resource_type_id = rt.id
-         left join aws_account aa on res.aws_account_id = aa.id
-where ia.private_ip = $1
-  and ia.not_before < $2
-  and (ia.not_after is null or ia.not_after > $2)
-`
+const resourceByPrivateIPQuery = `select * from get_resource_by_private_ip($1, $2)`
```
ensary, line 68: This massively improves readability, and is a good alternative to the other solution we discussed (sqlc - https://github.com/kyleconroy/sqlc). I wonder if we can combine both (sqlc also gives us golang stucts per row for "free")...
NIT: Did you consider calling via https://www.postgresql.org/docs/11/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-NAMED , I think it would make the queries inside the code even more readable.

db-migrations/000007_adding_functions.down.sql
rsoni-atlassian: Ah i missed it 👍

pkg/storage/db.go
rsoni-atlassian: Cool suggestion, i will look into incorporating them. Thank you!

rsoni-atlassian: On further review of sqlc, i think sqlc is powerful tool. Sqlc will definitely help us simplify our code by reducing boiler plate code. As followup i think we could track SQLC specific work as part of separate task as i do see changes being significant atleast in db.go code. Thoughts?

ensary: Of course, it's different scope. I mostly wondered how well would sqlc work with functions. Might want to spike this later.

pkg/storage/db_test.go
```diff
@@ -1291,15 +1249,7 @@ func TestGetARNIDAtTimeMoreThanOneChampions(t *testing.T) {
...
-		"yahoo.com",
-		"type",
-		"aid",
-		"region",
-		[]byte("{\"hi\":\"there3\"}"),
-		1)
-	rows2 := sqlmock.NewRows([]string{"aws_account_account",
+		"aws_account_account",
 		"owner_login",
 		"oener_email",
```
ensary, line 1254: Unrelated to your changes, but this looks like typo in column name. The test still works though, as we `Scan` not by name but by position

rsoni-atlassian: I looked further in the test and it seems like we have been using naming convention tablename_field. Test refactor task could be a good candidate to address this or leave as is.

db-migrations/000007_adding_functions.up.sql
```diff
@@ -0,0 +1,275 @@
...
+    FROM aws_private_ip_assignment ia
+             LEFT JOIN aws_resource res ON ia.aws_resource_id = res.id
+             LEFT JOIN aws_region ar ON res.aws_region_id = ar.id
+             LEFT JOIN aws_resource_type rt ON res.aws_resource_type_id = rt.id
+             LEFT JOIN aws_account aa ON res.aws_account_id = aa.id
+    WHERE ia.private_ip = pip
+      AND ia.not_before < ts
+      AND (ia.not_after IS NULL OR ia.not_after > ts);
+
+    SELECT temp_private_ip_table.id INTO var_account_id FROM temp_private_ip_table FETCH FIRST ROW ONLY;
```
dscgarcia, line 207: I'm new to sql functions so perhaps I'm misunderstanding something but does selecting only the first accountId mean we can safely make the assumption that if the temp_private_ip_table has multiple results that the accountId in each record will be the same as the first? 
I know it is rare for us to see multiple results in these queries, perhaps it is true that the accountId will be the same for all ip addresses, but I wanted to make sure.

rsoni-atlassian: This is a great question, i had similar concern while implementation. The existing implementation relied on time stamp `ts` and assumption/reality is there won't be any over lapping `start` and `end` ts. This gives us unique `aws_resource_id` which in turn point to unique `aws_account`.

However, we can certainly check for violation of uniqueness assumption and throw exception.

pkg/storage/db_test.go
ensary: I mean `oEner` instead of `owner` :)

db-migrations/000007_adding_functions.up.sql
rsoni-atlassian: Updated query to incorporate above regression. Also, added test data related to above scenario in latest commit.

PR #66, opened 2020-07-29
Report the state of database schema for GET version

ensary:
Introduce additional field in response for GET version that tells if the database schema is in "dirty" state after a failed migration.

### DIFF
--- a/api.yaml
+++ b/api.yaml
@@ -468,11 +468,11 @@ paths:
       description: "Allows to discover the current state of database schema for the specific deployment"
       responses:
         200:
-          description: "Schema Version"
+          description: "Schema State"
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/SchemaVersion"
+                $ref: "#/components/schemas/SchemaState"
       x-transportd:
         backend: app
         enabled:
@@ -811,6 +811,13 @@ components:
       properties:
         version:
           type: integer
+    SchemaState:
+      type: object
+      properties:
+        version:
+          type: integer
+        dirty:
+          type: boolean
     AccountOwner:
       type: object
       properties:
--- a/pkg/storage/schema.go
+++ b/pkg/storage/schema.go
@@ -112,7 +112,7 @@ func (sm *SchemaManager) migrateSchema(ctx context.Context, d migrationDirection
 	if err != nil {
 		return 0, err
 	}
-	version, _ , err := sm.GetSchemaVersion(ctx)
+	version, _, err := sm.GetSchemaVersion(ctx)
 	if err != nil {
 		return 0, err
 	}
--- a/pkg/storage/schema.go
+++ b/pkg/storage/schema.go
@@ -80,20 +80,20 @@ func (sm *SchemaManager) MigrateSchemaToVersion(ctx context.Context, version uin
 }
 
 // GetSchemaVersion retrieves the current version of database schema
-func (sm *SchemaManager) GetSchemaVersion(ctx context.Context) (uint, error) {
+func (sm *SchemaManager) GetSchemaVersion(ctx context.Context) (uint, bool, error) {
 	err := sm.EnsureConnected()
 	if err != nil {
-		return 0, err
+		return 0, false, err
 	}
-	v, _, err := sm.migrator.Version()
+	v, dirty, err := sm.migrator.Version()
 	if err == migrate.ErrNilVersion {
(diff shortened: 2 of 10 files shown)


pkg/storage/schema.go
```diff
@@ -80,20 +80,20 @@ func (sm *SchemaManager) MigrateSchemaToVersion(ctx context.Context, version uin
...
-	v, _, err := sm.migrator.Version()
+	v, dirty, err := sm.migrator.Version()
 	if err == migrate.ErrNilVersion {
 		// special handling for the version not being present
-		return 0, nil
+		return 0, dirty, nil
 	}
 	if err != nil {
-		return 0, err
+		return 0, dirty, err
```
rsoni-atlassian, line 94: question: is it safe assumption that any error is due to db in dirty state? could it be programming bug?

pkg/storage/schema.go
```diff
@@ -112,7 +112,7 @@ func (sm *SchemaManager) migrateSchema(ctx context.Context, d migrationDirection
 	if err != nil {
 		return 0, err
 	}
-	version, err := sm.GetSchemaVersion(ctx)
+	version, _, err := sm.GetSchemaVersion(ctx)
```
rsoni-atlassian, line 115: question: shall we include schema state in schema up and down api calls as well? or as part of response code? I am not sure if consumers would be able to self diagnose that schema is in bad state

api.yaml
```diff
@@ -811,6 +811,13 @@ components:
       properties:
         version:
           type: integer
+    SchemaState:
```
rsoni-atlassian, line 814: question: are there any consumers relying on this API? Is this breaking change?

pkg/storage/schema.go
ensary: No. Dirty state is not error, it is a valid state from migration library perspective when we ask it for the schema version.
Generally `dirty`, and any other return value (like schema version) is considered undefined for error cases and should not be checked by caller if `err` is set.  This is based on Go  general convention https://blog.golang.org/error-handling-and-go - for example, square root of negative number is not 0, it is an irrational number, but Go creators are fine with returning 0 as the sqrt of negative if `err` is not `nil`, as it is caller's responsibility to check for error and ignore any return values for failed calls.  
Whether we return true of false here should not matter, i.e. we could put `true` or `false` here and if the caller's behavior changed, it would be a bug in caller code. 
We *could* use non-boolean type here and have special `undefined` state in addition of true/false, but that would probably create more problems than it solves, as the library we are wrapping does use boolean. It also would arguably be Java-in-golang. One more approach could be to abuse the pointer to boolean, and return true/false for when we "know" the result, and nil pointer when we don't, but from what I understand this would not be idiomatic go either.

pkg/storage/schema.go
ensary: Consumers do not get new version of schema if the migration fails and schema is marked as dirty. They get non-200 response with specific error message.

api.yaml
ensary: No. This is endpoint that is known to be only used by people operating the service, so the only thing we possibly need to update is a runbook.

ensary: One place that needs to change is automation around moving db to specific version for integration tests of other services, but that needs to change after we merge.

pkg/storage/schema.go
ensary: On top of that, adding the field  would increase the scope of breaking change to scripts that use the API.

pkg/storage/schema.go
rsoni-atlassian: This is great, for a golang newbie thank you for sharing details

pkg/storage/schema.go
```diff
@@ -80,20 +80,20 @@ func (sm *SchemaManager) MigrateSchemaToVersion(ctx context.Context, version uin
 }
 
 // GetSchemaVersion retrieves the current version of database schema
-func (sm *SchemaManager) GetSchemaVersion(ctx context.Context) (uint, error) {
+func (sm *SchemaManager) GetSchemaVersion(ctx context.Context) (uint, bool, error) {
```
alecbert-atl, line 83: My only nit is that externally we now call this Schema State but internally we keep the function name GetSchemaVersion.

ensary: I agree,  we should probably later rename API endpoint and funcs

PR #71, opened 2020-08-10
Including resource-type as part of resource-id extraction from ARN

rsoni-atlassian:


db-migrations/000010_adding_unique_constraint_on_aws_resource_arnid_account_region.down.sql
```diff
@@ -0,0 +1,10 @@
+-- Removing unique constraint on aws_resource table for columns arn_id, aws_account_id, aws_region_id
+BEGIN;
+
+ALTER TABLE aws_resource
+    DROP CONSTRAINT IF EXISTS arn_account_region_id_unique;
+
+ALTER TABLE aws_resource
+    ADD CONSTRAINT aws_resource_arn_id_key UNIQUE (arn_id);
+
+COMMIT;
```
gcase555, line 10: nit: newline, ditto in the sql file below

(we should probably run a linter on our sql files at some point)

gcase555: (they probably should have had them but no biggie if not)

PR #73, opened 2020-08-20
SID-490 POC Integration tests

akhalina:
A basic integration test. Could use some feedback on:
1. Does it make sense to use the latest schema? And is there a better/easier way to get the database to the latest schema?
2. The client is generated based on the branch right now because if the api.yaml file isn't edited, the client will be generated the same as the master, but if it was edited then we should be able to test those changes. Does this make sense, or is it still better to generate from master?
3. I thought about using Rajiv's test data for postgres. Is this something we want, or is it better to just start with a blank database?

Locally you can use the `make local-integration` target for ease of use. This will clean up after the run, but if you plan on not using this target then ensure you clean up your docker containers between runs (especially postgres).


ensary:
> Does it make sense to use the latest schema? And is there a better/easier way to get the database to the latest schema? 
We should be able to run integration tests with several different schema versions. For now we could hard-code which ones do we want as the range. "Old" tests have an example of repeatedly  running same set of tests with different versions.
re: better/easier way - you can call binary for golang/migrate (they provide a docker image) instead of using the API if you find it easier to do. 

> The client is generated based on the branch right now because if the api.yaml file isn't edited, the client will be generated the same as the master, but if it was edited then we should be able to test those changes. Does this make sense, or is it still better to generate from master?
My first thought - both! As it allows to validate that changes on branch do not break things AND prove that the changes did not break backwards compatibility (via client bindings generated from "previous generation" api.yaml). 
The changes to API can add new endpoints/add optional input/output parameters/fields/make input fields optional, but should not break existing working API clients. 
We will have to jump through some hoops (get integration test code or possibly container image (?) from master to make sure it works server built from branch) but we already had couple of cases where we unintentionally broke API, so it is probably worth the investment. Not necessarily as the part of this PR, but the sooner the better.

> I thought about using Rajiv's test data for postgres. Is this something we want, or is it better to just start with a blank database?
Considering we need to support multiple schema versions (not yet supported in the test data for psql), and eventually ensure that the write endpoints function properly across the range of  schema versions - we can use a subset of existing test data, but we will have to add/check some more via API calls.

gcase555:
> Does it make sense to use the latest schema? And is there a better/easier way to get the database to the latest schema?

I concur with above, it depends on the integration test. Some tests we might want against most schema versions, some tests we will want to run against only specific schemas. The test setup should probably be responsible for doing this IMO either through the go migrate cli or through the asset inventory API

> The client is generated based on the branch right now because if the api.yaml file isn't edited, the client will be generated the same as the master, but if it was edited then we should be able to test those changes. Does this make sense, or is it still better to generate from master?

It makes sense to me to generate it from the current branch for testing purposes and I think is ideal.

> I thought about using Rajiv's test data for postgres. Is this something we want, or is it better to just start with a blank database?

The test data could be a decent starting point and I'd lean towards using at least a subset of it, if possible, instead of designing our own test data entirely in this case.

akhalina:
@gcase555 Just to clarify what you're saying about schema versions: it sounds like each test should handle moving through schemas themselves, meaning removing any schema movement from `TestMain` correct? Since each test could require testing a different subset of schemas.

.travis.yml
```diff
@@ -7,8 +7,11 @@ services:
...
 script:
+  - make generate-integration-client
   - travis_retry make dep
   - make lint
   - make test
+  - make integration
+  - make clean-integration
   - make coverage
-  - bash <(curl -s https://codecov.io/bash) -f .coverage/combined.cover.out
+  - bash <(curl -s https://codecov.io/bash) -f .coverage/combined.cover.out
```
gcase555, line 16: nit: missing newline

gcase555:
> it sounds like each test should handle moving through schemas themselves, meaning removing any schema movement from TestMain correct? Since each test could require testing a different subset of schemas.

It's probably better to have a generic function that you can pass a version into as a parameter, so the test setups can select a version. The potential local data loss needs to be considered when tests are downgrading the schema and could cause problems if this was run concurrently.

ensary:
> To confirm, in this case we would also have to copy and run the integration test files from master?

Yes, I think so. Otherwise we'd risk having unbuildable code. In fact, I think it would be beneficial to package latest "known good" integration test binary from master  as docker image. And use this image and possibly code built from branch simultaneously if needed (for example, if tests changed).

akhalina:
Update:
I added a helper function for stepping through schemas. It will still be up to the implementer of the test to make sure that they use this with caution especially when stepping down.

I added a make target for testing with master api.yaml and tests against local branch server, but I cannot add this to the pipeline until the master integration test setup is stable. This means adding a quick line in another PR.

Lastly, added version 1 schema compatible subset of our test data specifically for easier integration testing.

Please let me know if there are any concerns.

akhalina:
Notes on testing against master:
This involved retrieving a subset of files from master and overwriting them in the local branch. This is a little awkward, but relatively harmless.

Notes on writing integration tests:
The make target `generate-integration-client` is meant for use when writing tests locally. The client is always generated and should not be edited. Using this command allows you look at the available methods and to reference them in your IDE. All generated code will be contained in the `client` folder.

PR #77, opened 2020-08-27
Adding ability to handle ENI relationships

rsoni-atlassian:
### Overview

Please refer [SID-511](https://asecurityteam.atlassian.net/browse/SID-511)
Also, creating indexes on aws_resource_relationship table as per existing system we only maintain one entry for ADDED resource until we see a DELETED event. Without index we will have multiple entries which is not ideal.

### Testing

* Unit teting
* Ran migrations on local
* Manual end to end tests via POST /v1/cloud/change

### DIFF
--- a/db-migrations/000012_adding_indexes_on_aws_resource_relationship.up.sql
+++ b/db-migrations/000012_adding_indexes_on_aws_resource_relationship.up.sql
@@ -1,0 +1,7 @@
+-- Adding indexes on aws_resource_relationship
+BEGIN;
+
+CREATE UNIQUE INDEX IF NOT EXISTS aws_resource_relationship_idx_no_after ON aws_resource_relationship (not_before, arn_id, related_arn_id) WHERE not_after IS NULL;
+CREATE UNIQUE INDEX IF NOT EXISTS aws_resource_relationship_idx_no_before ON aws_resource_relationship (not_after, arn_id, related_arn_id) WHERE not_before IS NULL;
+
+COMMIT;
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -437,6 +437,16 @@ func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetCh
 					}
 				}
 			}
+			for _, res := range val.RelatedResources {
+				if strings.EqualFold(added, val.ChangeType) {
+					err = db.assignResourceRelationship(ctx, tx, arnID, res, cloudAssetChanges.ChangeTime)
+				} else {
+					err = db.releaseResourceRelationship(ctx, tx, arnID, res, cloudAssetChanges.ChangeTime)
+				}
+				if err != nil {
+					break
+				}
+			}
 			if err != nil {
 				break
 			}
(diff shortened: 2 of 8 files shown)


gcase555:
nit: no need to link to jira tickets from github since they are not accessible for OSS projects. It's locked down so outsiders can't access it but we try not to link to it in our OSS repo's.

db-migrations/000012_adding_indexes_on_aws_resource_relationship.up.sql
```diff
@@ -0,0 +1,7 @@
+-- Adding indexes on aws_resource_relationship
+BEGIN;
+
+CREATE UNIQUE INDEX IF NOT EXISTS aws_resource_relationship_idx_no_after ON aws_resource_relationship (not_before, arn_id, related_arn_id) WHERE not_after IS NULL;
```
gcase555, line 4: question: do we have performance concerns about adding this index without doing so `CONCURRENTLY`?

rsoni-atlassian: Not really, table is empty currently.

pkg/storage/db.go
```diff
@@ -1107,6 +1117,62 @@ func (db *DB) releasePublicIP(ctx context.Context, tx *sql.Tx, arnID string, ip
 	return err
 }
 
+func (db *DB) assignResourceRelationship(ctx context.Context, tx *sql.Tx, arnID string, resource string, when time.Time) error {
+	const assignResourceRelationshipQueryUpdate = `
+update aws_resource_relationship
+set not_before = $1
+where related_arn_id = $2
+  and not_before = to_timestamp(0)
```
gcase555, line 1125: question: I am slightly confused as to why this query has `to_timestamp(0)`? what is 0 in this context? is this needed?

pkg/storage/db.go
```diff
@@ -437,6 +437,16 @@ func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetCh
 					}
 				}
 			}
+			for _, res := range val.RelatedResources {
```
gcase555, line 440: info: non-blocking concern. This function is getting rather large at 60 lines and It has quite a bit of loops that could be separated into their own functions.

Scope wise I don't think it needs to change here since it was like this already but for readabilities sake we should probably break this function up in the near future.

rsoni-atlassian: I agree, there is lot of room for refactor and improving code quality in db.go

pkg/storage/db.go
rsoni-atlassian: Basically `to_timestamp(0)` is to convert a Unix epoch value to a PostgreSQL timestamp. This implementation is similar to other implementation like `PrivateIpAddress` and `PublicIpAddress`, here we are only over writing not_before if it is set to default `to_timestamp(0)`

pkg/storage/db.go
```diff
@@ -1107,6 +1117,62 @@ func (db *DB) releasePublicIP(ctx context.Context, tx *sql.Tx, arnID string, ip
...
+	_, err = tx.ExecContext(ctx, assignResourceRelationshipQueryInsert, when, resource, arnID)
+	return err
+}
+
+func (db *DB) releaseResourceRelationship(ctx context.Context, tx *sql.Tx, arnID string, resource string, when time.Time) error {
+	const releaseResourceRelationshipQueryUpdate = `
+update aws_resource_relationship
+set not_after=$1
+where related_arn_id = $2
+  and arn_id = $3;`
```
dscgarcia, line 1154: I don't know how often this happens (if at all) but if we were to have to two records with the same `arn_id` and `related_arn_id` then this update statement would update both, so to be safe we might want to add `and not_after is null` to the where clause section. We have something similar here (line 1098) https://github.com/asecurityteam/asset-inventory-api/pull/77/files#diff-106eb1177cdcbe545695a4fdec2e2a32R1098

rsoni-atlassian: Great point @dscgarcia i doubt we will hit this scenario but completely agree addition of `and not_after is null` is good to be safe.

PR #78, opened 2020-08-27
Create new fetchByResourceID endpoint

sydneyteh96:
A new endpoint is created and is similar to fetchByARNID endpoint.


pkg/storage/db.go
```diff
@@ -898,6 +898,111 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 	return cloudAssetDetails, err
 }
 
+// FetchByResourceID gets the assets who have resource ID at the specified time
+func (db *DB) FetchByResourceID(ctx context.Context, when time.Time, resID string) ([]domain.CloudAssetDetails, error) {
```
alecbert-atl, line 902: Question/Suggestion: Having a separate handler mostly makes sense to me so we can have different validation messages, but a quick comparison of FetchByResourceID and FetchByARNID shows the functions are about equivalent besides having resID and arnID switched. 

Is there anything stopping you from having the handler call FetchByARNID in the background?

sydneyteh96: I think the name "FetchByResourceID" explains what it does because the function gets resources by resource ID. Also, the name matches with handler name. Other than this, they are mostly the same.

akhalina: I think it would be fair to not copy the arnid methods and to instead rename them to resource by id which is more clear. Then have both the "arnid" endpoint and "resid" endpoint point to the same method in the backend. This is fine because the purpose of this new endpoint is for clarity to the external user. Having the same method in the backend will not affect what users expectations from these endpoints, ie they are currently supposed function the same with arnid being deprecated.

sydneyteh96: I think this is a good suggestion. I will change the endpoints and have both endpoints point to the same method. So "fetch by arn id" will still work after deprecated.

api.yaml
```diff
@@ -381,6 +381,65 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
+  /v1/cloud/resid/{resid}:
```
akhalina, line 384: nit/opinion: From what I can tell we don't usually abbreviate our endpoints. I think it would be more clear to use something like:  `/v1/cloud/resourceid/{resourceid}`

main.go
```diff
@@ -137,7 +137,8 @@ func (c *component) New(ctx context.Context, conf *config) (func(context.Context
 		"insert":                     serverfull.NewFunction(insert.Handle),
 		"fetchByIP":                  serverfull.NewFunction(fetchByIP.Handle),
 		"fetchByHostname":            serverfull.NewFunction(fetchByHostname.Handle),
-		"fetchByArnID":               serverfull.NewFunction(fetchByArnID.Handle),
+		"fetchByArnID":               serverfull.NewFunction(fetchByResourceID.Handle),
```
ensary, line 140: 👍 , we will be able to remove ByArnID later once we are sure the client is updated.

gcase555:
I will address some of the minor PR feedback here on Monday and also get this re-reviewed and merged.

api.yaml
gcase555: fixed

pkg/storage/db.go
gcase555: I believe this has been addressed.

akhalina: Indeed it has.

akhalina (dismissed):

gcase555:
FWIW I have tested that this works for resource id based lookups. It appears arn based lookups are broken, I am trying to fix that now.

gcase555:
Ok I fixed it, I did need to change the name of the parameter for `arnid` based lookups to `resourceid` but this parameter is in the URL path, so consumers should not be able to tell the difference. The resource id is passed in as part of the URL the same as the arn id was, its just named differently behind the scenes.

gcase555:
I tested this by checking the database for resources that existed and doing 2 similar queries to make sure these resources existed and returned the same data.

Both work as expected
`curl http://localhost:8080/v1/cloud/resourceid/i-00000000000000002?time='2019-07-25T08:53:41.318Z'`

`curl http://localhost:8080/v1/cloud/arnid/i-00000000000000002?time='2019-07-25T08:53:41.318Z'`

api.yaml
```diff
@@ -362,11 +362,70 @@ paths:
...
+              "status":
+              #! if eq .Response.Body.errorType "InvalidInput" !# 400,
+              #! else !#
+              #! if eq .Response.Body.errorType "NotFound" !# 404,
+              #! else !# 500,
+              #! end !#
+              #! end !#
+              "bodyPassthrough": true
+            }
+  /v1/cloud/resourceid/{resourceid}:
```
rsoni-atlassian, line 384: nit: path as per REST `/v1/cloud/resource/{resourceid}`  It looks like current implementation pattern is similar to arnid.

api.yaml
```diff
@@ -322,13 +322,13 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
-  /v1/cloud/arnid/{arnid}:
+  /v1/cloud/arnid/{resourceid}:
     get:
       summary: "Retrieve a cloud asset at a point in time by ARN ID"
```
rsoni-atlassian, line 327: nit: update from `ARN ID` to `Resource ID`

api.yaml
```diff
@@ -322,13 +322,13 @@ paths:
               #! end !#
               "bodyPassthrough": true
             }
-  /v1/cloud/arnid/{arnid}:
+  /v1/cloud/arnid/{resourceid}:
```
rsoni-atlassian, line 325: Question: Why not use `deprecated` field in swagger for now to communicate with consumers?

rsoni-atlassian: Additional note:

I am fine with this change, however from consumers perspective this change is confusing for following reasons
* API path is `/v1/cloud/arnid/` looking at this information as consumer i expect `arnid` lookup
* If we are communicating with consumers to pass `resourceid` then i would prefer to communicate new endpoint `/v1/cloud/resourceid/{resourceid}` and mark existing endpoint `/v1/cloud/arnid/{arnid}` as deprecated

api.yaml
gcase555: do you mean the comment here? I can change that

api.yaml
gcase555: FWIW we are not conforming to REST standards. 

In this case, we want to be explicit that the endpoint is for resource id based lookups so I am leaning towards keeping it the same.

api.yaml
gcase555: It's resourceid because we wanted to use the same function behind the scenes, I needed to rename the parameter to match the format and parameter of the new function.

I should clarify that no new parameter is being named here, consumer behavior is unchanged. It's just the pathname. Consumers won't have to change anything here.

gcase555: Also, we are fast following this ticket with the deprecation so this is a temporary state that should not exist for more than a few days.

api.yaml
gcase555: `resource` is too generic as we have separate endpoints for each lookup type, in this case, this is a resource id based lookup. If we will have a generic `resource` lookup, then the parameter should allow them to choose between ip, hostname or resource id but that would be a significant change.

api.yaml
gcase555: I'll do this in a separate PR

rsoni-atlassian: yep, text in summary

api.yaml
rsoni-atlassian: sounds good, i remember we had similar discussion offline.

api.yaml
rsoni-atlassian: ok, lets leave as is.

PR #79, opened 2020-08-31
Replace inner joins with left joins on resource to champion

akhalina:
Replaced inner joins with left joins so we do not return an empty set when there are no account owners. This was tested with schema version 12 and 13 with and without an account owner, and with and without champions. Both versions work with no champions, only 13 works with no account owner. Also needed to make some strings nullable because of the null fields that are now returned when there is no account owner. Did a little research, and it seems pointers are the easiest way to work with null strings returned by the sql library we are using.

This is my first time doing a migration, so let me know if this is incorrect!

### DIFF
--- a/db-migrations/000013_updating_joins.up.sql
+++ b/db-migrations/000013_updating_joins.up.sql
@@ -1,4 +1,4 @@
--- schema changes to create functions for APIs
+-- update functions to use left joins
 BEGIN;
 
 CREATE OR REPLACE FUNCTION get_resource_by_arn_id(aid VARCHAR, ts TIMESTAMP)
--- a/db-migrations/000013_updating_joins.up.sql
+++ b/db-migrations/000013_updating_joins.up.sql
@@ -1,0 +1,321 @@
+-- schema changes to create functions for APIs
+BEGIN;
+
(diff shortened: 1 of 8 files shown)


db-migrations/000013_updating_joins.up.sql
```diff
@@ -0,0 +1,321 @@
...
+                          SELECT distinct iwres.aws_account_id, f.*
+                          FROM wres iwres,
+                               LATERAL get_owner_and_champions_by_account_id(iwres.aws_account_id) f
+                      ) b
+                      ON wres.account = b.t_account;
+END;
+$$
+    LANGUAGE 'plpgsql';
+
+CREATE OR REPLACE FUNCTION get_owner_and_champions_by_account_id(id INTEGER)
```
rsoni-atlassian, line 76: Question: Are there any implementation changes in this function?

db-migrations/000013_updating_joins.up.sql
```diff
@@ -0,0 +1,321 @@
+-- schema changes to create functions for APIs
```
rsoni-atlassian, line 1: nit: please update description elaborating changes

db-migrations/000013_updating_joins.up.sql
akhalina: That's a good catch! I'll remove this.

PR #81, opened 2020-08-31
Fix error with ARNID lookup

akhalina:
I made the mistake in my previous PR of not properly testing the ARNID lookup. I needed to use pointers properly for that one as well. This time I tested the ip, hostname (these use the same method), and arnid lookups locally.

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -837,10 +837,10 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 		var publicIPAddress sql.NullString
 		var hostname sql.NullString
 		var metaBytes []byte
-		var chLogin string
-		var chEmail string
-		var chName string
-		var chValid bool
+		var chLogin *string
+		var chEmail *string
+		var chName *string
+		var chValid *bool
 		if err = rows.Scan(&privateIPAddress, &publicIPAddress, &hostname, &asset.ResourceType, &asset.AccountID,
 			&asset.Region, &metaBytes, &accountID, &account.AccountID, &account.Owner.Login, &account.Owner.Email,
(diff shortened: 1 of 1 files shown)


pkg/storage/db.go
```diff
@@ -864,11 +864,13 @@ func (db *DB) FetchByARNID(ctx context.Context, when time.Time, arnID string) ([
 			hasTag = true
 		}
 
-		tempChampionsMap[chLogin+chEmail] = domain.Person{
-			Login: &chLogin,
-			Email: &chEmail,
-			Name:  &chName,
-			Valid: &chValid,
+		if chLogin != nil {
+			tempChampionsMap[*chLogin] = domain.Person{
```
akhalina, line 868: I'm not 100% sure about this change. Apparently the login is unique, and I assume the same champion would not appear for the same resource twice. But please correct me if I'm wrong.

gcase555: It's technically possible an account owner could also be an account champion although I don't know of any cases in our data where this is true.

akhalina: In this case the old way this was done would still cause issues correct?

gcase555: Cause what specific issues?

gcase555: I am looking at how the IP based queries are executed and the logic is slightly different which is interesting, is it possible the issue exists there as well?

akhalina: I guess I just want to know why we had the map key as chLogin+chEmail before? Is there some risk to just using the login?

akhalina: Sorry, to clarify there was no issue that I know of in the previous code. I can add a check for nil on both the login and email, I just don't know if that is necessary/preferred.

gcase555: Login is guaranteed to be unique so I am unsure why they would combine the two for a key also

PR #83, opened 2020-09-03
Fix null pointer error

akhalina:
I thought I had tested all possible null variations but it seems I missed one. This would be great to get through today so that on-call doesn't get any more alerts on this.

I'd like to go back and clean this code up at some point, but for now this will do.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ test:
 # Generate the client used for integration tests. For local development.
 generate-integration-client:
 	docker run --rm \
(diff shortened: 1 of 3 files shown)


akhalina:
Additionally I ran into an issue where the latest version of the OpenAPI generator isn't generating models anymore and thus is breaking our pipeline. I am hoping they fix this, but in the mean time I've anchored a working version.

rsoni-atlassian (approved):

dscgarcia (approved):

PR #87, opened 2020-09-09
Integration test for Cloud Changes endpoint - happy path

ensary:
First take on integration tests with "real-ish" data.
* Factor out helper functions/sample data
* Check schema version at startup, run all tests for a range of schemas we support - we might want to re-run from scratch with empty db and certain schema version to raise confidence later.
* Happy path is validated via matching lookups for all resources that were in the change under test. 
* Some of the tests for input validation are disabled/commented out as they triggered yet-unknown bugs. I will create ticket per bug that will have "uncomment test X" as part of DoD.

WIP - fixing timing/race condition in container startup related to psql, should be a trivial fix.


api.yaml
```diff
@@ -879,13 +879,21 @@ components:
             $ref: "#/components/schemas/CloudAssetDetails"
     CloudAssets:
```
ensary, line 880: This is a fix for a bug where generated swagger bindings could not parse response from Asset-Inventory due to mismatch between the code and api.yaml. This changes api.yaml to match code, as existing clients clearly are able to parse responses, so we are using code as source of truth.

integration/tests/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+	openapi "github.com/asecurityteam/asset-inventory-api/client"
+)
+
+var assetInventoryAPI *openapi.APIClient //this contains pre-configured API client used by all fixtures
+
+func TestMain(m *testing.M) {
+	config := openapi.NewConfiguration()
+	appURL := os.Getenv("AIA_APP_URL")
+	config.BasePath = appURL
+	_, config.Debug = os.LookupEnv("AIA_INTEGRATION_DEBUG")
```
ensary, line 20: This is extremely valuable when debugging integration tests as it sends all interactions with api to stderr verbatim.

integration/tests/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		panic(err)
+	}
+	//run all know tests with all supported schema versions
+	for v := int32(12); v <= maxSchema; v++ {
+		err := setSchemaVersion(v)
+		if err != nil {
+			panic(fmt.Errorf("error migrating database schema %#v", err))
+		}
+		res += m.Run()
```
ensary, line 36: This runs tests discovered from all files in this package, hence 'main_test.go' as the filename.

Makefile
```diff
@@ -23,6 +23,9 @@ test:
         -w "$(DIR)" \
         asecurityteam/sdcli:v1 go test
 
+psql-console:
```
ensary, line 26: Runs psql console if you local stack is up.

integration/tests/healthcheck_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHealthcheck(t *testing.T) {
+	tt := []struct {
+		Name string
+	}{
+		{
+			Name: "Health Check",
```
rsoni-atlassian, line 18: Question: Do we see need to add more test cases here?

integration/tests/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"os"
+	"testing"
+
+	openapi "github.com/asecurityteam/asset-inventory-api/client"
+)
+
+var assetInventoryAPI *openapi.APIClient //this contains pre-configured API client used by all fixtures
+
+func TestMain(m *testing.M) {
```
rsoni-atlassian, line 16: nit: rename to TestSchemaVersion or something better?

integration/tests/schema_helpers_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	openapi "github.com/asecurityteam/asset-inventory-api/client"
+)
+
+var schemaVersion int32  //current schema version
+var maxSchema int32 = 14 // TODO: extrapolate this somewhere?
```
rsoni-atlassian, line 13: nit: rename to maxSchemaVersion

integration/tests/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+	appURL := os.Getenv("AIA_APP_URL")
+	config.BasePath = appURL
+	_, config.Debug = os.LookupEnv("AIA_INTEGRATION_DEBUG")
+	assetInventoryAPI = openapi.NewAPIClient(config)
+	ctx := context.Background()
+	schemaVersion = getSchemaVersion(ctx, assetInventoryAPI.DefaultApi)
+	res := 0
+
+	//run all know tests with all supported schema versions
+	for v := int32(12); v <= maxSchema; v++ {
```
rsoni-atlassian, line 27: Question: Does the order of integration test execution matter? I think we may face unexpected failure if we are running this test in conjunction with other tests.

integration/tests/sample_data_test.go
```diff
@@ -0,0 +1,74 @@
...
+package tests
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	openapi "github.com/asecurityteam/asset-inventory-api/client"
+)
+
```
rsoni-atlassian, line 12: Question: What are your thoughts on additional test data? Are you expecting this implementation to grow?

integration/tests/healthcheck_test.go
ensary: Out of scope for my planned work. This is refactoring of what @akhalina did in here previous spike, unchanged.  The expectation is - there is a file per api endpoint, and this file is supposed to cover healthchecks across different schema versions.

integration/tests/main_test.go
ensary: https://golang.org/pkg/testing/#hdr-Main . This is golang entrypoint for test run. Think main(). Magic name, so can not rename :(

integration/tests/main_test.go
ensary: Schema changes definitely will result in failure if used with same database concurrently. So order of fixtures (i.e. schema generations) matter, or rather - only one can be used with same db. Order of tests inside fixture should not (all tests are run for every fixture). This way we can run multiple db instances concurrently with different fixtures (version of schema), but not with different schemas on same db concurrently.

integration/tests/schema_helpers_test.go
ensary: This is copy/paste from what was master. Git for some reason does not reflects it. I'll change the name :)

ensary: Done, also switched to const.

integration/tests/healthcheck_test.go
ensary: At least that's the pattern I followed so far

akhalina: Not really sure what we can add here. This was meant as more of a sample of what tests should look like.

ensary: This one is useful as it is run for every "fixture" (i.e. full test suite run with different schema version). And if healthcheck fails, the rest of the tests are pretty much guaranteed to fail.

integration/tests/sample_data_test.go
ensary: We also will definitely need test data for resource types other than EC2 Instance.

integration/tests/healthcheck_test.go
akhalina: I agree, definitely useful to get a general sense of if the service in the test is cooperating.

rsoni-atlassian: Sounds good, since tests were parametrized that made me think we may want to add more test parameters in future

integration/tests/sample_data_test.go
rsoni-atlassian: ok, just wanted to make sure i follow pattern

rsoni-atlassian (approved):

akhalina (approved):

alecbert-atl (approved):

PR #88, opened 2020-09-09
Add validation for aws account input

akhalina:
Added a separate object schema to validate input of owners and champions (there is a ticket to re-combine when we are ready).

I would love any input on whether this is too strict, or even too loose. 

This has been tested locally, we have a ticket to address integration testing on this endpoint.

I will need to followup with updating api.yaml on BB as well.

### DIFF
--- a/api.yaml
+++ b/api.yaml
@@ -784,7 +784,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/AccountOwner"
+              $ref: "#/components/schemas/SetAccountOwner"
       responses:
         201:
           description: "An AWS account with its owner and champions are inserted or updated"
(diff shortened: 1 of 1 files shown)


ensary (approved):

api.yaml
```diff
@@ -987,6 +987,39 @@ components:
...
         valid:
           type: boolean
+    SetAccountOwner:
+      type: object
+      properties:
+        accountId:
+          type: string
+          pattern: ^(\d{12})$
+        name:
+          type: string
```
dscgarcia, line 997: Are we adding `name` to the AccountOwner object as well?

akhalina: Oops, no. Good catch! I'll remove that.

dscgarcia (approved):

rsoni-atlassian (approved):

PR #89, opened 2020-09-09
updating min schema version to be current

gcase555:
As explained in our runbook, the minimum default schema version should be updated after a new version has been successfully deployed and migrated.

### DIFF
--- a/integration/tests/client_test.go
+++ b/integration/tests/client_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 var assetInventoryAPI *openapi.APIClient
-var schemaVersion int32 = 14
+var schemaVersion int32 = 13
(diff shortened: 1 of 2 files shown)


gcase555:
annoyance:
Schema needs to be updated in three places:
- the makefile for headless runs
- the min version in the schema code
- the integration tests

I was surprised when I had to do this, it wasn't obvious I needed to change them in all three places.  We might consider having the integration tests get the max version automatically from the code in a follow-up PR.

akhalina:
The numbers from the code weren't up to date at the time the integration code was written, but I agree we should definitely combine as much of these as we can.

gcase555:
tasks:
- [ ] rebase from master
- [ ] document updating the default schema so its 2 versions behind

PR #96, opened 2020-09-11
Add validation for Account ID and re-enable/add integration tests

ensary:
Re-uses the validator @akhalina wrote for other endpoints and adds tests.

### DIFF
--- a/api.yaml
+++ b/api.yaml
@@ -828,7 +828,7 @@ components:
         resourceType:
           type: string
         accountId:
-          type: string
+          $ref: "#/components/schemas/AWSAccountID"
         region:
           type: string
         arn:
@@ -910,7 +910,7 @@ components:
         resourceType:
           type: string
         accountId:
-          type: string
+          $ref: "#/components/schemas/AWSAccountID"
         region:
           type: string
         arn:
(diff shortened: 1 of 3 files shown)


api.yaml
```diff
@@ -972,7 +972,7 @@ components:
     AccountOwner:
       type: object
       properties:
-        accountId:
+        accountId: # TODO add specific type once we are sure we return it correctly
```
ensary, line 975: I am not 100% sure we stopped returning null here, so not adding the validation here.

sydneyteh96 (dismissed):

rsoni-atlassian (approved):

akhalina (approved):

akhalina:
My bad, the build is failing.

integration/tests/main_test.go
```diff
@@ -23,6 +23,7 @@ func TestMain(m *testing.M) {
 	appURL := os.Getenv("AIA_APP_URL")
 	config.BasePath = appURL
 	_, config.Debug = os.LookupEnv("AIA_INTEGRATION_DEBUG")
+	config.Debug = true
```
ensary, line 26: This seems to completely remove any flakes :(

ensary: I assume forcing I/O after every request prevents one or more race conditions from happening. Let's see if this is true...

api.yaml
akhalina: Yes, I definitely wouldn't touch the account owner stuff for lookups yet.

akhalina (approved):

rsoni-atlassian (approved):

PR #98, opened 2020-09-11
Fix transaction getting poisoned

akhalina:
Ensuring that when an error happens for a cloud change POST we return an error right away so that the transaction isn't hanging around with an error causing subsequent queries to fail. This does mean that if any sql query fails on a request the whole request fails. 

Note: This was not and will not be poisoning following requests. The problem was with queries in the same transaction. When one query fails and breaks the transaction, the following queries in the transaction cannot run properly and we get a bunch of error logs. Not dire, but not ideal either.

I tried to find a way to write an integration test to make sure this fails gracefully, but luckily I couldn't find a way to replicate this problem with all of our new validations.

 (Also lint doesn't like "cancelled" as a spelling)

### DIFF
--- a/pkg/storage/db.go
+++ b/pkg/storage/db.go
@@ -416,50 +416,7 @@ func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetCh
 		return err
 	}
 	if err = db.ensureResourceExists(ctx, cloudAssetChanges, tx); err == nil {
-		arnID := resIDFromARN(cloudAssetChanges.ARN)
-		for _, val := range cloudAssetChanges.Changes {
-			for _, ip := range val.PrivateIPAddresses {
-				if strings.EqualFold(added, val.ChangeType) {
-					err = db.assignPrivateIP(ctx, tx, arnID, ip, cloudAssetChanges.ChangeTime)
-				} else {
-					err = db.releasePrivateIP(ctx, tx, arnID, ip, cloudAssetChanges.ChangeTime)
-				}
-				if err != nil {
-					break
-				}
-			}
-			if err != nil {
-				break
-			}
-			for _, ip := range val.PublicIPAddresses {
-				for _, hostname := range val.Hostnames { //TODO look very closely into awsconfig-tranformerd logic for this
-					if strings.EqualFold(added, val.ChangeType) {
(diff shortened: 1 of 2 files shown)


ensary (approved):

alecbert-atl (dismissed):

pkg/storage/db.go
```diff
@@ -428,15 +428,24 @@ func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetCh
...
 						err = db.assignPublicIP(ctx, tx, arnID, ip, hostname, cloudAssetChanges.ChangeTime)
 					} else {
 						err = db.releasePublicIP(ctx, tx, arnID, ip, hostname, cloudAssetChanges.ChangeTime)
 					}
+					if err != nil {
+						break
+					}
 				}
 			}
+			if err != nil {
```
rsoni-atlassian, line 446: nit: i dont think this is necessary.

rsoni-atlassian (dismissed):

pkg/storage/db.go
akhalina: Could you clarify why?

rsoni-atlassian: I was referring more to pattern of using `break` in case of error. There are multiple ways in which we can make this pattern better:

1. Extract for loops into functions and return error or empty values in error case
2. Use Goto keyword, not clean but better than current pattern
3. Use Label to exit out of inner and outer loops

akhalina: That's fair, it's not the prettiest looking code. I am skeptical about using goto and labels, those feel very similar to me and I believe are generally discouraged for readability's sake. Splitting out the for loop into a function is an option, I can try to do that!

pkg/storage/db.go
```diff
@@ -428,15 +428,24 @@ func (db *DB) StoreV2(ctx context.Context, cloudAssetChanges domain.CloudAssetCh
 					break
 				}
 			}
+			if err != nil {
+				break
+			}
 			for _, ip := range val.PublicIPAddresses {
```
akhalina, line 434: I missed one for loop...

ensary (approved):

rsoni-atlassian (approved):

PR #99, opened 2020-09-11
Remove old schema and related code

ensary:
* Added migration that drops all the old tables (v1) completely
* Removed any endpoints valid only with old schema in place (partitions and backfill) and related code
* Removed any code dispatching based on old/new schema version - this includes dual-writes and reads-based-on-version
* Fixed all the tests to pass with these changes, including integration w/o old tables in place


db-migrations/000015_drop_legacy_schema.up.sql
```diff
@@ -0,0 +1,7 @@
+-- WARNING!!! this is not possible to roll back w/o back-fill of data from new schema or some other source
+-- no transaction as it would be supermassive
+DROP TABLE IF EXISTS aws_events_ips_hostnames; -- this automatically gets rid of partitions
```
rsoni-atlassian, line 3: Question: I am bit concerned about our ability to not restore data, could we copy content from tables to other tables *_backup? And drop them after we are 100% sure

rsoni-atlassian (commented):
Suggestion: I think we would also need to regenerate scripts under `sample-data`

db-migrations/000015_drop_legacy_schema.up.sql
ensary: We can keep one of the snapshots. Copying tables will take a ton of time. One more thing - we stopped dual-writes (i.e. writing to the tables we are dropping) couple months ago, so the data in old tables is very stale anyway.

rsoni-atlassian (dismissed):

pkg/storage/db.go
```diff
@@ -536,140 +299,26 @@ func resIDFromARN(ARN string) string {
...
-	}
-	var assets []domain.CloudAssetDetails
-	if ver < ReadsFromNewSchemaVersion {
-		sqlstmt := fmt.Sprintf(latestStatusQuery, `aws_hostnames_hostname`)
-		assets, err = db.runQuery(ctx, sqlstmt, hostname, when)
-	} else {
-		assets, err = db.runFetchByIPQuery(ctx, false, resourceByHostnameQuery, hostname, when)
-	}
-	return assets, err
+	return db.runFetchByIPQuery(ctx, false, resourceByHostnameQuery, hostname, when)
```
alecbert-atl, line 309: Nit: while we're in here, is it possible to refactor `runFetchByIPQuery`'s name since we're also searching by hostname with it?

pkg/storage/db.go
```diff
@@ -536,140 +299,26 @@ func resIDFromARN(ARN string) string {
...
-	// this lib won't give back the last INSERTed row ID, so we don't bother with `RETURNING ...`
-	// See https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql
-	_, err := tx.ExecContext(ctx, fmt.Sprintf(`INSERT INTO %s VALUES ($1, $2, $3, $4, $5, $6)`, tableAWSEventsIPSHostnames), timestamp, isPublic, isJoin, resourceID, ipAddress, hostname) // nolint
-	return err
-}
-
 // FetchAll gets all the assets present at the specified time
 func (db *DB) FetchAll(ctx context.Context, when time.Time, count uint, offset uint, typeFilter string) ([]domain.CloudAssetDetails, error) {
-	return db.runQuery(ctx, bulkResourcesQuery, when, typeFilter, count, offset)
+	return nil, errors.New("bulk export API is not available")
```
alecbert-atl, line 304: Question/Nit: Is this because the bulk query relied on the old schema layout? 

In terms of nit, I think we've got a Pollinator check that hits the bulk endpoint (I'm not sure if we own it though) from our Splunk logs. If we're going to be turning the endpoint off like this it'd be nice to have a followup to stop pinging the endpoint. 

Also I see that the bulkResourcesQuery is still in the const list at the top of db.go. I think it should be taken out if this was the only place that used it.

ensary: It is *not* supposed to work (at least it does not return any sensible results as dual-writes stopped long time ago). If there is pollinator check - we should stop it. 
We are about to re-write this code path to use new schema.
re: Query - I will drop it, thank you!

ensary: re: Not supposed to work - it has such poor performance that running it can result in an outage. We should have added code to error immediately long ago.

pkg/storage/db.go
ensary: Fixed

alecbert-atl (dismissed):
Thank you for the changes.

alecbert-atl (approved):

rsoni-atlassian (approved):

sydneyteh96 (approved):

pkg/domain/storage.go
```diff
@@ -46,43 +30,11 @@ type CloudAllAssetsByTimeFetcher interface {
...
-}
-
-func (e NotFoundPartition) Error() string {
-	return fmt.Sprintf("A partition does not exist for the requested name, %s", e.Name)
-}
-
 // EventExportHandler handles exporting a single event during export
 type EventExportHandler interface {
 	Handle(changes CloudAssetChanges) error
 }
```
dscgarcia, line 36: If we are getting rid of `events_local_backfill.go` can we eliminate this Handler as well? or is it used by something else?

integration/tests/lookup_by_resourceid_test.go
```diff
@@ -70,5 +70,5 @@ func TestLookupByResourceID(t *testing.T) {
 			})
 	}
 
-	RawUrlFollowupTests(t, "/v1/cloud/resourceid/" + resId, tsDuring)
+	RawUrlFollowupTests(t, "/v1/cloud/resourceid/"+resId, tsDuring)
```
gcase555, line 73: info: I am guessing this is some kind of auto-formatting? is it preferred we don't put a space after the string when concatenating? If so, TIL I will try to be mindful of that.

gcase555 (approved):

dscgarcia (approved):

integration/tests/lookup_by_resourceid_test.go
ensary: Goimports did this :)

pkg/domain/storage.go
ensary: Will follow-up with removal, good catch.

PR #100, opened 2020-09-14
Account for resource type being overwritten by broken test

ensary:
https://github.com/asecurityteam/asset-inventory-api/blob/master/integration/tests/cloudchanges_test.go#L107 is where resource type is owerwritten due to lack of validation (in a race-condition-prone fashion :(). 
We can not rely on proper resource type to be present in response for other tests.
This should unbreak master branch. Follow-up PR to re-enable master run coming next.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ script:
   - make integration
   - make coverage
   - bash <(curl -s https://codecov.io/bash) -f .coverage/combined.cover.out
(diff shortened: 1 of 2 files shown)


alecbert-atl (approved):
Unfortunate

dscgarcia (approved):

akhalina (approved):

PR #104, opened 2020-09-16
minimum schema version updated

gcase555:
The goal of this pull request is to update the min schema version to something closer to current than 6(which is way behind).

The integration tests broke in `master-integration` but will likely not work until I merge in this change and add them back to the travis builds.

### DIFF
--- a/pkg/storage/schema.go
+++ b/pkg/storage/schema.go
@@ -20,8 +20,6 @@ const (
 const (
 	// EmptySchemaVersion Version of database schema that cleans the database completely. Use cautiously!
 	EmptySchemaVersion uint = 0
-	// MinimumSchemaVersion Lowest version of database schema current code is able to handle
-	MinimumSchemaVersion uint = 1
 	// M1SchemaVersion Version of database schema with performance optimizations (M1) that allows back-fill to work
 	M1SchemaVersion uint = 2
 	// DualWritesSchemaVersion Lowest version of database schema that supports dual-writes (legacy and M1)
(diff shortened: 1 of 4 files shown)


pkg/storage/schema.go
```diff
@@ -30,6 +28,8 @@ const (
 	ReadsFromNewSchemaVersion uint = 4
 	// NewSchemaOnlyVersion Lowest version that stops dual-writes in preparation to drop old schema
 	NewSchemaOnlyVersion uint = 6
+	// MinimumSchemaVersion Lowest version of database schema current code is able to handle
+	MinimumSchemaVersion uint = 13
```
alecbert-atl, line 32: On our BB side @ensary notes that we can't rollback once we migrate to schema version 15. Is this the same thing as needing our minimum schema version to be 15? I'm not clear on what happens if we migrate an environment to step up to 15 and then do a redeploy a week later, does this property being 13 mean on redeploy AI-API will start at 13?

Also I know this PR wasn't made to address it, but I see that we have properties around dual writes above, should these be removed now that release v0.40.0 of AI-API has taken those out?

gcase555: I should probably make this schema version 15 if we can't go backward from that

gcase555: ok so 15 fails the builds, looking into why now

dscgarcia: I think we need to also update the min version in main_test.go https://github.com/asecurityteam/asset-inventory-api/blob/master/integration/tests/main_test.go#L34, I didn't think about that with the PR to remove all of the old schema code. I'm surprised my PR went through, I didn't update this and tests still showed passing...

gcase555: Ill do that now

gcase555: we'll likely keep 13 here.

dscgarcia (approved):

akhalina (approved):

PR #106, opened 2020-09-17
Allow multiple arn id, as long as the region and/or owners are different.

akhalina:
Two things this PR doesn't change:
1. Relationship arnid use
2. Returning multiple assets for the resource id lookup

About #2: We hadn't fully discussed whether we want to be able to return multiple assets for a resource id lookup. If we want this I can add to another PR. Something to note -- when I make this change locally we return over 3000 records for our "Valid" test because we use the same timestamps for many entries and multiple owner/champion combos. I assume this wouldn't be the case in real life, but if this is a concern I can also address this in another ticket.

### DIFF
--- a/integration/tests/cloudchanges_test.go
+++ b/integration/tests/cloudchanges_test.go
@@ -4,6 +4,7 @@ package tests
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


gcase555:
🤔  I am debating if it is ever valid to receive multiple assets for a point in time lookup. If a resource id is unique, shouldn't there only be one resource returned when a specific timestamp is given?

akhalina:
The point of this PR in general is that we don't have a guarantee of a unique resource id. From what I understand CLBs specifically break this rule. Our database is now set up such that our unique constraint in on a combo of resource id, account id, and region. For this reason we can't guarantee there is only one resource per resource id.

ensary (dismissed):

gcase555 (approved):

ensary (approved):

gcase555:
thank you for increasing coverage here

gcase555 (approved):

PR #109, opened 2020-09-18
Sid 654/multiple resource params

alecbert-atl:
This uses the new custom template function "reduce" to transform path parameters into a valid resourceId. This is our workaround for swagger interpreting `/` in a path parameter as being part of the path itself.

Currently working on issue where our openapi client generator does not interpret YAML anchors despite them being valid YAML.

### DIFF
--- a/gateway.Dockerfile
+++ b/gateway.Dockerfile
@@ -1,3 +1,3 @@
-FROM serverfull-gateway:local
+FROM asecurityteam/serverfull-gateway:v1.0.5
 COPY api.yaml .
 ENV TRANSPORTD_OPENAPI_SPECIFICATION_FILE="api.yaml"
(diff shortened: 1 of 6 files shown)


gateway.Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM asecurityteam/serverfull-gateway
+FROM serverfull-gateway:local
```
ensary, line 1: Don't forget to remove this pls!

ensary (approved):

ensary (dismissed):

alecbert-atl:
😦  The Open API Spec validator doesn't like the YAML anchors despite them working when I build our gateway locally. Looking in to what I can do that isn't turning off the validation or adding more copy/paste

sydneyteh96 (approved):

nyanshak (approved):

PR #110, opened 2020-09-18
Add resourcetype validation

dscgarcia:
Adds validation for resourceType on cloud_insert. Updated integration tests for testing on resource type. Updated unit test with valid data as well.
I had to temporarily disable `make master-integration` because the changes to integration tests is not compatible with master branch (sample-data had an invalid resource type as passing). Will follow up with a quick PR to re-enable master-integration step in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ script:
   - make coverage
   - bash <(curl -s https://codecov.io/bash) -f .coverage/combined.cover.out
   - make integration
-  - make master-integration
+#  - make master-integration
--- a/pkg/handlers/v1/cloud_insert_test.go
+++ b/pkg/handlers/v1/cloud_insert_test.go
@@ -72,3 +72,14 @@ func TestInsertStorage(t *testing.T) {
 	e := newInsertHandler(storage).Handle(context.Background(), validInsertInput())
 	assert.Nil(t, e)
 }
+
+func TestInsertInvalidResourceType(t *testing.T) {
+	input := CloudAssetChanges{
+		ResourceType: "MS:Windows:2000",
+	}
+	e := newInsertHandler(nil).Handle(context.Background(), input)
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -13,4 +13,4 @@ script:
   - make coverage
   - bash <(curl -s https://codecov.io/bash) -f .coverage/combined.cover.out
   - make integration
-  - make master-integration
+#  - make master-integration
```
dscgarcia, line 16: will follow up quickly with another pr to re-enable this step

pkg/handlers/v1/cloud_insert_test.go
```diff
@@ -24,7 +24,7 @@ func validInsertInput() CloudAssetChanges {
 	return CloudAssetChanges{
 		ChangeTime:   time.Now().Format(time.RFC3339Nano),
 		ARN:          "cloud-resource-arn",
-		ResourceType: "cloud-resource-type",
+		ResourceType: "AWS::EC2::Instance",
```
dscgarcia, line 27: needed to update resourceType for unit tests to pass

.travis.yml
gcase555: info: we may consider removing this build step entirely if we feel it continues to be problematic. Its been more noise than signal for me lately as well.

ensary: We need to think about alternatives re: not breaking API. As the test from master looks like the only thing that does it (even if poorly).

ensary (approved):

akhalina:
It could be helpful to also add an enum to api.yaml for this particular field. It can be useful as a reference for any users of our api (especially in the returned object).

alecbert-atl (approved):
I was actually just going to ask if you could add a test like `TestInsertInvalidResourceType` so this is great!

akhalina (approved):

PR #111, opened 2020-09-18
WIP: adding migrations for removing redundant records

gcase555:
Think this query is pretty close, I am loading up some data locally to test this more thoroughly but feedback is welcome. 

See Issue SID-678 for more details. We essentially have many redundant records for this resource id, our plan is to delete the records for this resource ID and run a backfill to repopulate these private ip assignments correctly now the update statement has been modified to probably prevent these redundant records from coming back.


db-migrations/000016_remove_duplicate_ip_assignments.up.sql
```diff
@@ -0,0 +1,16 @@
+-- select records we will remove into the new backup table in case we need to recover them
+select aws_private_ip_assignment.id,
```
rsoni-atlassian, line 2: Question: What is the process to clean backup_* table?

Suggestion: It would be good to backup whole database/table and restore, you may see better performance (depending on data size) as well.

gcase555: We will just delete the backup_ table after a certain amount of time and verifying the data looks correct.

This could be overkill, I do have the records exported up to a 30k+ long .sql file that is full of insert statements. I hesitate to check it in as a migration if we feel its too large.

gcase555: We could also just use a snapshot if we really had to rollback but I doubt it would go that far, I would just run the insert statements and that would re-populate the records as it was

db-migrations/000016_remove_duplicate_ip_assignments.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS backup_aws_private_ip_assignment;
```
gcase555, line 1: question: should I add an insert statement here to get the records back into the original table before dropping this?

gcase555: (I am guessing yes but want to be sure of approach here before going deeper on it)

rsoni-atlassian: yep, that works.

db-migrations/000016_remove_duplicate_ip_assignments.up.sql
gcase555: in terms of record size, it would be inserting around 31,000 records so I don't believe the performance hit would be to significant.

db-migrations/000016_remove_duplicate_ip_assignments.down.sql
gcase555: TODO: insert records back into table from the backup table before dropping it

db-migrations/000016_remove_duplicate_ip_assignments.up.sql
gcase555: discussed in slack, we will do a follow-up task in jira to remove the backup table

rsoni-atlassian: Yep, that was my concern as well leading me to lean more towards snapshot.

gcase555:
moved to bitbucket

